PR #5, opened 2022-11-14
Story - IO 164 - project card models and store

EngbergJonas:
- Interface for project card
- Store and slice for project card
- Folder restructure so that views are only main views
- Added unit tests next to their components being tested
- Added index files for each component folder for cleaner imports
- Added mock files for reacti18-next and project card
- Changed test-utils file name to testUtils to support the camelCase rule


kgunga (approved):
👍

PR #6, opened 2022-11-15
Story - IO-167 - Hankekortti connect to local API

EngbergJonas:
- axios
- axios-interceptor, 
- projectCard service
- projectCard thunk for calling the service and updating the redux state
- Dotenv to webpack to handle .env files


README.md
```diff
@@ -10,8 +11,15 @@
     - [State and storage](#state-and-storage)
     - [Routing](#routing)
     - [Testing](#testing)
+    - [Localization](#localization)
     - [Collection of scripts](#collection-of-scripts)
 
+### Setup
+
+1. `npm run create-env` to create your own development and production .env files
```
kgunga, line 19: use yarn create-env instead

README.md
```diff
@@ -10,8 +11,15 @@
...
+2. `yarn install`
+3. `npm start`
```
kgunga, line 21: use yarn start instead. we should use yarn tool everywhere

kgunga (changes requested):

kgunga (changes requested):
I think we should use react-scripts instead of using directly webpack. react-scripts already wraps webpack.

README.md
EngbergJonas: Resolved.

README.md
EngbergJonas: Resolved.

kgunga (approved):
Looks good

PR #10, opened 2022-12-06
Bug IO-201 project card header editable

EngbergJonas:
- Saving of header form
- Moved some stuff to utils
- Tests for saving of header form

### DIFF
--- a/src/components/ProjectCard/ProjectCardBasicsForm.tsx
+++ b/src/components/ProjectCard/ProjectCardBasicsForm.tsx
@@ -1,6 +1,6 @@
 import useProjectCardBasicsForm from '@/hooks/useProjectCardBasicsForm';
 import { useAppDispatch, useAppSelector } from '@/hooks/common';
-import { IProjectCardBasicsForm } from '@/interfaces/formInterfaces';
+import { IAppForms, IProjectCardBasicsForm } from '@/interfaces/formInterfaces';
 import { FC } from 'react';
 import { SubmitHandler } from 'react-hook-form';
 import { FormFieldCreator } from '../shared';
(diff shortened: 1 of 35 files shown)


src/mocks/mockUser.ts
```diff
@@ -1,7 +1,7 @@
 import { IUser } from '@/interfaces/userInterfaces';
 
 const mockUser: IUser = {
-  id: '9d6a0854-a784-44b0-ad35-ca5e8b8f0e90',
+  id: '9990de74-96ac-4896-bd5f-f4bb506da38f',
```
kgunga, line 4: Why this value has changed? Checking the Add to favourite star breaks the save operation

kgunga (changes requested):

src/components/ProjectCard/ProjectCardBasicsForm.tsx
```diff
@@ -51,7 +38,7 @@ const ProjectCardBasicsForm: FC = () => {
         <div className="basic-info-form">
           <FormFieldCreator form={formFields} />
         </div>
-        <Button type="submit">{t('send')}</Button>
+        <Button type="submit">Tallenna perustiedot</Button>
```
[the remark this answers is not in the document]
EngbergJonas: I thought I would just hardcode this value, since this button isn't going to exist after autosaving is enabled. Should we still move it to fi.json?

src/mocks/mockUser.ts
EngbergJonas: This whole operation needs to be changed to work with the current person from our API /persons. I forgot to create a thunk to fetch the user. The operation breaks, since the mock user doesn't exist in our database.

kgunga (approved):
👍

PR #11, opened 2022-12-08
Story/io 197/planning list view

EngbergJonas:
- Initial planning list view
- Planning list tables and other necessary components
- Navigate to project card from planning table


kgunga (approved):
👍

PR #20, opened 2023-01-25
Story/io 218/search or create hashtags

EngbergJonas:
- HashTags form new design

- New components:
  - ProjectHashTags (hashtags as a whole)
  - NewHashTagsForm (POST new hashtags, only usable by admins in the future)
  - HashTagSearch (search all hashtags from listSlice)
  - HashTagsContainer (for displaying hashtags)
  
- New List to listSlice:
  - hashTags
  
- Tests for hashtags


src/hooks/useHashTagsForm.ts
```diff
@@ -0,0 +1,26 @@
+import { IHashTagsForm } from '@/interfaces/formInterfaces';
+import { useMemo } from 'react';
+import { useForm } from 'react-hook-form';
+
+const useHashTagsValues = () => {
```
kgunga, line 5: I don't understand this how it is supposed to work. It is a hook which never changes the value of formValues

EngbergJonas: This form was created into its own function, because all other forms have also been sepparated into their own functions. Since react-hook-form is a hook, creating a sepparate function for retrieving the form values also needs to be a hook. That's the only reason.

kgunga: alright

kgunga (approved):
👍

PR #24, opened 2023-02-06
Story/io 244/handling popular hashtags

Git-Uzair:
- Added popular hashtags to the hashtags modal
- Created new IHashTag response interface for the new payload from hashtags endpoint
- Fixed a bug where same hashtags could be added to a project multiple times in the modal
- Altered code to comply with the new hashtags endpoint response where needed
- Wrote test for new changes


src/reducers/listsSlice.ts
```diff
@@ -35,7 +38,7 @@ export interface IListState {
   division: Array<IListItem>;
   subDivision: Array<IListItem>;
   responsibleZone: Array<IListItem>;
-  hashTags: Array<IListItem>;
+  hashTags: IHashTagsResponse;
```
EngbergJonas, line 41: We need to create a new store for hashtags, since the hashtags are no longer of type IListItem.

src/mocks/mockLists.ts
```diff
@@ -294,33 +295,43 @@ export const mockResponsibleZones: { data: Array<IListItem> } = {
...
-    {
-      id: '816cc173-6340-45ed-9b49-4b4976b2a48b',
-      value: 'hulevesi',
-    },
-    {
-      id: '8346d18f-f836-4179-b0e7-74ee43cb65b0',
-      value: 'aukio/tori',
-    },
-    {
+export const mockHashTags: { data: IHashTagsResponse } = {
```
EngbergJonas, line 298: Move this to a new file mockHashTags.ts when creating a new slice.

PR #25, opened 2023-02-07
Story/io 18/project phase to planning list

EngbergJonas:
- close phase dialog when patching phase
- remove notification when patching phase from planning list
- update list view phase when patching phase from project card form
- removed unnecessary debugs from projectCardBasics.test.tsx
- removed timeout extensions from projectCardBasics.test.tsx


kgunga (approved):
👍

PR #34, opened 2023-03-08
Story/io 156/show category in planning list view

EngbergJonas:
- Added category tag to planning table project header 
- Removed icons not to be used in MVP

PR #35, opened 2023-03-09
Story/io 264/plan list show current view

EngbergJonas:
- Changes to planning info panel for planning list view


EngbergJonas:
https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/34

kgunga (approved):
I think navigating back should be done from the up pointing arrow but this is current design

PR #37, opened 2023-03-10
Story/io 7/color code project cells

EngbergJonas:
- Add an active color for project cells that have either planning, construction or both during that year
- Create project cells from the correct values
- Only allow editing of cell values that are active


src/components/Planning/PlanningGroupsTable/PlanningGroupsTableRow.tsx
```diff
@@ -3,24 +3,49 @@ import { planProjectValues } from '@/mocks/common';
...
+      type: getCellType(year),
+      objectKey: key,
+      projectId: project.id,
+    };
+  };
+
+  const cells = [];
 
-interface IPlanningProjectsTableProps {
+  for (const [key, value] of Object.entries(project)) {
```
kgunga, line 35: return Object.entries(project).map or flatMap

src/components/Planning/PlanningGroupsTable/PlanningGroupsTableRow.tsx
```diff
@@ -29,13 +54,15 @@ interface IPlanningProjectsTableProps {
 /**
  * We're only mapping the project name here for now since the values aren't yet implemented
  */
-const PlanningGroupsTableRow: FC<IPlanningProjectsTableProps> = ({
+const PlanningGroupsTableRow: FC<IPlanningGroupsTableRowProps> = ({
```
kgunga, line 57: onClick={(e) => onProjectMenuClick(project.id, e as unknown as MouseEvent)} move this anonymous function into const onClick = useCallback((e) => ...., [project])

src/components/Planning/PlanningGroupsTable/ProjectCell.tsx
```diff
@@ -0,0 +1,49 @@
...
+        onBlur={handleBlur}
+        onFocus={handleFocus}
+        onChange={handleChange}
+        disabled={type === 'none'}
+      />
+    </td>
+  );
+};
+
+export default ProjectCell;
```
kgunga, line 49: Component receives properties so wrap with memo

src/utils/common.ts
```diff
@@ -126,3 +126,21 @@ export const classesToOptions = (classes: Array<IClass>): Array<IOption> =>
...
+ * @returns boolean
+ */
+export const isInYearRange = (
+  year: number,
+  startDate: string | undefined,
+  endDate: string | undefined,
+): boolean => {
+  const startYear = parseInt(startDate?.split('.').pop() || '0');
+  const endYear = parseInt(endDate?.split('.').pop() || '0');
+  return startYear && endYear ? year >= startYear && year <= endYear : false;
```
kgunga, line 145: can we use here moment to check is year between start and end years?

kgunga (changes requested):

src/utils/common.ts
EngbergJonas: moment is added is the next feature https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/38.

src/components/Planning/PlanningGroupsTable/PlanningGroupsTableRow.tsx
EngbergJonas: This behaviour changes in https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/38, could it be fixed there?

PR #42, opened 2023-04-17
Story/io 282/empty select field

EngbergJonas:
- New empty selection button for SelectField component that looks like the one HDS provides for MultiSelect fields


kgunga (approved):
👍

PR #43, opened 2023-04-17
bug/io-238/previous-search-criteria-not-cleared

EngbergJonas:
-set multi-select options to match the redux storeFormValues when the state changes in useSearchForm()


kgunga (approved):
👍

PR #49, opened 2023-05-25
Fixed styling issue with side navigation on ProjectBasicsForm

Git-Uzair:
Fixed styling issue with the sidepanel on Project Basics form

- set width to 100% for the side-nav class
- set pw link text font size to 14px and color to black
- moved styles to classes instead of inline

- Fixed Uusi menu header to match designs


kgunga (approved):
👍

PR #53, opened 2023-06-08
Story/io 311/returning to planning view

EngbergJonas:
- new planning slice
- moved all variables created and managed by usePlanningRows()-hook to new planning slice
- added ScrollHandler component to keep track of scroll height when navigating back and forth between ProjectBasicsView and other views
- added a box shadow to planning toolbar to reflect the design
- moved finance-update event to App.tsx to keep the updated class, location and group values at all times


kgunga (approved):
👍

PR #55, opened 2023-06-13
Story/io 308/planning view table improvements

EngbergJonas:
- highlight columns in PlanningView where cursor is pointing to help identify which year a cell belongs to
- tooltip when hovering the monthly timeline that shows the exact start and end dates for the hovered timeline
- sticky positioning to planning summary table so that the year summary is always shown when scrolling the y-axis
- sticky positioning to planning table head cells so that the rows name is always visible when scrolling the x-axis
- subClasses that have "suurpiiri" in their name now act as a district when selected


kgunga (approved):
👍

PR #59, opened 2023-06-21
Story/io 316/direct non programmed projects to project form from search results

EngbergJonas:
- Added a tag for projects that are not programmed
- Navigate not programmed projects to project form instead of planning view 
- **Needs to be merged with the backend fix: https://github.com/City-of-Helsinki/infraohjelmointi-api/tree/fix/search-results-projects-response**


kgunga (approved):
👍

PR #63, opened 2023-07-06
Story/io 315/coordinator view

EngbergJonas:
**This branch needs to be merged after the backend changes to the search results path, since the path now needs to include search params instead of nested pathing, https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/55.**

The branch features the frame of the coordination view. The following main features were added:

General changes:
- PlanningView.tsx now uses url search params instead of nested routes, since nested routes would not work with having the same type of PlanningRows in multiple depths.
- PlanningRowTypes are now camel cased to better reflect the url search params used
- CSS-changes to the new and existing levels to also work with the coordination view level depths

planningSlice.tsx:
- new mode property to determine the PlanningView.tsx component is in coordination or planning mode
- additional selections to include the new levels: selectedCollectiveSubLevel, selectedSubLevelDistrict, selectedOtherClassification

classSlice.tsx
- separated the old classes into a planning object and added a new coordination object with additional class types

locationSlice.tsx
- separated the old locations into a planning object and added a new coordination object with only the districts

usePlanningRows.tsx
- does nothing if planningMode === 'coordination'
- refactored away some functions into planningUtils.ts that are now also used by the new useCoordinationRows.tsx
- removed creating links since we now use search params
- refactored route logic into new usePlanningRoutes.tsx hook

**new** useCoordinationRows.tsx:
- does nothing if planningMode === 'planning'
- handles the building of coordination rows for the coordination view (mode) of PlanningView.tsx

**new** usePlanningRoutes.tsx:
- handles route related changes to the planning view and 
- populates planningSlice with selected items if they are found in the search params
- changes mode of the planning view based on the route used /planning or /coordination


Git-Uzair (approved):
Approved

PR #66, opened 2023-07-24
io-299/ Finance update when changing dates in project card

Git-Uzair:
- Added logic to move finances correctly when changing dates in project card

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
@@ -66,8 +66,12 @@ const ProjectForm = () => {
       if (isDirty) {
         dispatch(setIsSaving(true));
 
-        const data: IProjectRequest = dirtyFieldsToRequestObject(dirtyFields, form as IAppForms);
-
+        const data: IProjectRequest = dirtyFieldsToRequestObject(
+          dirtyFields,
+          form as IAppForms,
+          project,
+        );
+        // syncFinancesIfNeeded(project, data);
         // Patch project
         if (project?.id && projectMode === 'edit') {
           try {
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
@@ -71,7 +71,7 @@ const ProjectForm = () => {
           form as IAppForms,
           project,
         );
-        // syncFinancesIfNeeded(project, data);
+
         // Patch project
         if (project?.id && projectMode === 'edit') {
           try {
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -4,7 +4,12 @@ import { IAppForms, FormValueType } from '@/interfaces/formInterfaces';
 import { TFunction } from 'i18next';
 import { getYear, updateYear } from './dates';
 import _ from 'lodash';
-import { IProjectRequest } from '@/interfaces/projectInterfaces';
+import {
+  IProject,
+  IProjectEstDates,
+  IProjectFinances,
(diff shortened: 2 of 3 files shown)


src/utils/common.ts
```diff
@@ -129,10 +137,133 @@ export const dirtyFieldsToRequestObject = (dirtyFields: object, form: IAppForms)
...
+    budgetObject: IProjectFinances,
+    projectTimelineDates: IProjectEstDates,
+  ): IProjectFinances => {
+    /**
+     * Moves the provided finance value to its new finance location and sets current to null
+     * @param newFieldYear year for the new finance field which needs to be updated
+     * @param financeValue value which will be summed in the new year
+     * @param currentFinanceFieldName finance field name which the value belongs to
+     */
+    const updateFinanceValue = (
```
kgunga, line 165: Move this function outside the adjustBudgets function and provider parameter for adjustedBudgetObjet. Currently it is very confusing implementation

src/utils/common.ts
```diff
@@ -129,10 +137,133 @@ export const dirtyFieldsToRequestObject = (dirtyFields: object, form: IAppForms)
...
+    const constructionEndYear = estConstructionEnd ? getYear(estConstructionEnd) : null;
+
+    const adjustedBudgetObject: IProjectFinances = {
+      ...budgetObject,
+    };
+    // looping over all budgetfields in the project
+    let fieldName: keyof IProjectFinances;
+    for (fieldName in adjustedBudgetObject) {
+      if (fieldName !== 'year') {
+        // current budget field year, extract from field name
```
kgunga, line 204: current budget field year is year itself in adjustedBudgetObject. Here logic tries to read years in future. Write a better comment here, please

kgunga (changes requested):

src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
```diff
@@ -66,8 +66,12 @@ const ProjectForm = () => {
...
         dispatch(setIsSaving(true));
 
-        const data: IProjectRequest = dirtyFieldsToRequestObject(dirtyFields, form as IAppForms);
-
+        const data: IProjectRequest = dirtyFieldsToRequestObject(
+          dirtyFields,
+          form as IAppForms,
+          project,
+        );
+        // syncFinancesIfNeeded(project, data);
```
kgunga, line 74: Remove commented line or is this forgotten to be uncommented?

src/utils/common.ts
```diff
@@ -129,10 +137,137 @@ export const dirtyFieldsToRequestObject = (dirtyFields: object, form: IAppForms)
...
+      8: 'preliminaryCurrentYearPlus8',
+      9: 'preliminaryCurrentYearPlus9',
+      10: 'preliminaryCurrentYearPlus10',
+    };
+    const newFinanceFieldName: keyof IProjectFinances =
+      yearToBudgetKey[newFieldYear - budgetObject.year];
+    (budgetObject[newFinanceFieldName] as string) = String(
+      safeParseInt(budgetObject[newFinanceFieldName]) + financeValue,
+    );
+    (budgetObject[currentFinanceFieldName] as null) = null;
```
kgunga, line 187: Why budgetObject[currentFinanceFieldName] casted as null and set its value to null? Is intension to delete the value or just set to null?

kgunga (changes requested):

src/utils/common.ts
Git-Uzair: To set it to null, the field has to be casted as null or else it throws an error. 
Intention is to delete the finance value from that field, it is set to null to make sure it is not visible in planning view timeline.

kgunga (approved):
👍

PR #68, opened 2023-08-01
Fix/projects missing planning view

Git-Uzair:
- Fixed an issue where projects directly under a suurpiiri were not visible in the planning view

### DIFF
--- a/src/utils/planningRowUtils.ts
+++ b/src/utils/planningRowUtils.ts
@@ -21,12 +21,12 @@ export const getSortedProjects = (
   type: PlanningRowType,
   projects: Array<IProject>,
   name: string,
+  districtsForSubClass?: Array<IClass>,
 ) => {
   let projectList: Array<IProject> = [];
   switch (type) {
     case 'class':
     case 'subClass':
-    case 'subClassDistrict':
       projectList = projects
         .filter(
           (p) =>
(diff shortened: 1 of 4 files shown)


src/utils/planningRowUtils.ts
```diff
@@ -21,12 +21,12 @@ export const getSortedProjects = (
   type: PlanningRowType,
   projects: Array<IProject>,
   name: string,
+  districtsForSubClass?: Array<IClass>,
```
EngbergJonas, line 24: getSortedProjects could be made more readable now that it has more params and is pretty hard to read, here's a suggestion:

```
/**
 * Filters projects under an IPlanningRow if the correct conditions are met.
 *
 * @param id the id of the current item that the row is being created for
 * @param type IPlanningRowType
 * @param projects projects to map
 * @param districtsForSubClass when the type is a subClassDistrict then we will not render districts and we need this to see if the project would otherwise have appeared under the district
 * @returns
 */
export const getSortedProjects = (
  id: string,
  type: PlanningRowType,
  projects: Array<IProject>,
  districtsForSubClass?: Array<IClass>,
) => {
  const getProjectsForClasses = () =>
    projects.filter((p) => !p.projectGroup && !p.projectLocation && p.projectClass === id);

  const getProjectsForSubClassDistrict = () => {
    const projectHasDistrictOrNoLocation = (p: IProject) =>
      districtsForSubClass?.some((d) => d.id === p.projectLocation) || !p.projectLocation;

    return projects.filter(
      (p) => !p.projectGroup && projectHasDistrictOrNoLocation(p) && p.projectClass === id,
    );
  };

  const getProjectsForGroup = () =>
    projects.filter((p) => p.projectGroup).filter((p) => p.projectGroup === id);

  const getProjectsForLocation = () =>
    projects.filter((p) => !p.projectGroup).filter((p) => p.projectLocation === id);

  switch (type) {
    case 'class':
    case 'subClass':
      return sortByName(getProjectsForClasses()) as Array<IProject>;
    case 'subClassDistrict':
      return sortByName(getProjectsForSubClassDistrict()) as Array<IProject>;
    case 'group':
      return sortByName(getProjectsForGroup()) as Array<IProject>;
    case 'district':
    case 'division':
      return sortByName(getProjectsForLocation()) as Array<IProject>;
  }

  return [];
};
```

EngbergJonas (changes requested):

EngbergJonas (approved):

PR #80, opened 2023-09-06
Story/io 322/authentication

EngbergJonas:
- New react-oidc-context ja oidc-client-ts libraries
- Fetching UI and API access tokens
- Authentication when starting the application if the user is not authenticated
- Fetch app data only when a user is authenticated


kgunga (approved):
👍

PR #82, opened 2023-09-08
Fixed timeline phase change issue when no dates are set and only year…

Git-Uzair:
- Added fix for timeline phase change when  project est dates are null


kgunga (approved):
👍

PR #85, opened 2023-09-14
Story/io 367/adding groups to coordination

Git-Uzair:
- Added groups in coordination view
- Created new thunks for coordination  groups
- Modified group state in redux


src/App.tsx
```diff
@@ -57,6 +57,8 @@ const App: FC = () => {
       dispatch(getGroupsThunk()),
       dispatch(getPlanningClassesThunk()),
       dispatch(getPlanningLocationsThunk()),
+      dispatch(getGroupsThunk()),
```
EngbergJonas, line 60: getGroupsThunk is now called twice, it should be named to getPlanningGroupsThunk and only called ones. The action should be named to 'groups/getAllPlanning'.

EngbergJonas (approved):

PR #86, opened 2023-09-15
Story/io 368/forced to frame view finances

EngbergJonas:
- Additional coordination finance levels (locations)
- Forced to frame view finances
- Add TA-muutos to frame budget
- Redirect the authenticated user back to the initially opened page after authentication
- Remove TA-muutos from groups and planning view


kgunga (approved):
👍

PR #87, opened 2023-09-15
story/io-370/hide-non-programmed-projects-after-project-update

EngbergJonas:
- filters out the project from planning rows if the project-update event returns a project where programmed: false


kgunga (approved):
👍

PR #88, opened 2023-09-19
story/io-371/project-update-event-updates-every-users-project

EngbergJonas:
- don't update the selectedProject if the project-update event sends another project than the one being edited


kgunga (approved):
👍

PR #90, opened 2023-09-21
story/io 56/10 year view management

EngbergJonas:
- New YearSelector component
- Get financial data for previous and future years for classes, locations, group and projects
- Change all instances of Promise().then() to try catch blocks


kgunga (approved):
👍

PR #96, opened 2023-10-09
Story/io 373/enable negative values ta muutos

markspl:
With this, users can

- add negative numbers
- send value updating request only when the input field includes a value when closing
- remove text from input field (before it did not allow users to remove, replaced empty input field with "0"

Issue:
- when clicking the button (number) which opens the input field, it will be empty _if_ the field was left empty previously
  - um, huh? What do you mean? How to test this?
    1) Open a TA field
    2) Clear it and close the input field (by clicking on somewhere)
    3) Re-open the same input field
    4) Because it was left empty, it stays empty

  - What should happen?
    - When user clears the input field, closes it, and re-opens it, it should show the same value as the button text does.


saijaseppa (approved):
👍🏻

millaeerola (approved):

PR #99, opened 2023-10-11
Hard coded user roles

JenniLehtonen:
I added all the possible roles but others than admin and programmer are commented out as they weren't used at the moment.


src/reducers/authSlice.ts
```diff
@@ -22,55 +22,146 @@ const initialState: IUserState = {
...
+  "name": "sg_kymp_sso_io_koordinaattorit" as UserRole,
+  "display_name": "sg_KYMP_sso_IO_Koordinaattorit",
+}
+const projectManager = {
+  "id": '31f86f09-b674-4c1d-81db-6d5fe2e587f9',
+  "name": 'sg_kymp_sso_io_projektipaallikot' as UserRole,
+  "display_name": 'sg_KYMP_sso_IO_Projektipaallikot',
+}*/
+const admin = {
+  "id": 'test-admin-role',
```
JenniLehtonen, line 37: This id should propably be something else?

millaeerola:
Loging issues seem to be solved now so this should no longer be needed.

PR #100, opened 2023-10-13
Story/io 384/enable negative values ta muutos

markspl:
- show correct number on input value when it is opened
- don't allow to make request with negative numbers


millaeerola (approved):
Good job!

PR #101, opened 2023-10-24
Bug/fix excel import script

millaeerola:
- Modified planning view components so that we can show "other classification" level classes in the the planning view as well
- Some search functionality tests were skipped for no reason -> they are't anymore


markspl (approved):
Looks good 👍

PR #104, opened 2023-10-31
IO 57: title changes related to year change

millaeerola:
- Modified getPlanningRowTitle function to return the title based on current year and not based on index (for example: previously "kuluva TA" title was always given to the selected year and now it's give to the actual current year)
- Titles should now change automatically when the year changes
- Ticket: [https://helsinkisolutionoffice.atlassian.net/browse/IO-57](https://helsinkisolutionoffice.atlassian.net/browse/IO-57)
- "TP" title was not added cause "tilinpäätös" functionality is not implemented yet


saijaseppa (approved):
Good solution!

PR #107, opened 2023-11-06
update calc and send int value to backend

markspl:
- Update calc to check that frame budget value is not negative
- Send number and not string to backend


millaeerola (approved):
Nice!

PR #110, opened 2023-11-07
Release 7.11.2023

millaeerola:
- Excel import script changes and UI changes to correct the class hierarchy and TAE data
- TA-muutos field allows negative values
- Planning table headers changes according to year
- SAP data is shown for projects that have a SAP id


markspl (approved):
Looks good! 😎

PR #111, opened 2023-11-09
Fix: 404 error page after login

millaeerola:
- user is now redirected to planning view after login instead of the 404 error page
- Ticket: [https://futurice.atlassian.net/browse/HKISD-7](https://futurice.atlassian.net/browse/HKISD-7)


markspl (approved):
🎉

PR #112, opened 2023-11-09
Fix: author of a note

millaeerola:
- Note author was "undefined undefined" before. Now it should be fixed with new notes. Some old notes might still have "undefined undefined" cause they either didn't have an author when read from excels or the author info didn't come from PW
- Ticket: [https://futurice.atlassian.net/browse/HKISD-4](https://futurice.atlassian.net/browse/HKISD-4)


saijaseppa (approved):
Good job!

PR #120, opened 2023-11-21
Fix/project card responsible persons menus

millaeerola:
- Removed the list of persons that comes from PW from the person programming list, cause we want to implement that within the tool and not fetch it from PW
- Added sorting for names, so that they are now sorted by last name


markspl (approved):
💎

markspl (approved):

PR #123, opened 2023-11-29
Fix/remove category k5

markspl:
- fix translations: option.KX was not translated correctly on the dropdown list because value to find translation includes a dot (e.g. "K5.1")
- remove other mentions about K5 (should not be visible and subcategories (K5.1, ...) should be used)

TODO after merging to dev/main:
- remove unnecessary category K5 from database

https://futurice.atlassian.net/browse/HKISD-11


millaeerola (approved):
👍

PR #125, opened 2023-12-04
Fix/cell update loading

markspl:
- add loading component when updating cell data on Ohjelmointi- and Koordinointinäkymä -views
- loading component was cropped behind form's leftest column and top row
  - change z value bigger

image

### DIFF
--- a/src/components/Loader/styles.css
+++ b/src/components/Loader/styles.css
@@ -1,5 +1,5 @@
 .loader-container {
-  @apply fixed z-[9] -mt-20 h-full w-full overflow-hidden;
+  @apply fixed z-[110] -mt-20 h-full w-full overflow-hidden;
 }
 
 .loader-wrapper {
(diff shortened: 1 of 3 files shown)


JenniLehtonen (approved):
Works nicely! :smile: I tried to update cell data in programming view but no input field was opened even though I tried to click like every number there. I was able to edit cell data in coordinator view though and the loading spinner worked well there. Is the problem with the programmer view a known issue or is something weird just going on in my computer?

millaeerola:
> Works nicely! 😄 I tried to update cell data in programming view but no input field was opened even though I tried to click like every number there. I was able to edit cell data in coordinator view though and the loading spinner worked well there. Is the problem with the programmer view a known issue or is something weird just going on in my computer?

Summary class budgets are only editable in coordinator view. In programming view, only project budgets should be editable
Screenshot 2023-12-05 at 8 56 22

PR #126, opened 2023-12-08
Fix/new projects missing

markspl:
- update outdated browserlist
- removed extra setIsSaving dispatch
- require district form input before creating/saving a project
- add missing translations

Would be merged after adding missing districts (https://futurice.atlassian.net/browse/HKISD-3)


millaeerola (approved):
Looks good! 👍  Waiting for district options PR to be merged first

PR #127, opened 2023-12-15
Make group creating possible even if there's no subclasses

saijaseppa:
### Description 
Problem was that if class didn't have any subclasses, it was not possible to add group there. Subclass was required to choose and choosing was impossible when class didn't have any subclasses. 
Also when creating new group form was not suggesting any projects to be added in group if there was not subclass. 

This was fixed now so, that new groups can be added without defining subclass only if class doesn't have subclasses. If class has subclasses that is mandatory to define. 

### How to test
Create new group to path 803 Kadut ja liikenneväylät -> Yhteishankkeet väyläviraston kanssa. There is no subclasses and you can still create a new group.


src/components/Planning/GroupDialog/GroupDialog.tsx
```diff
@@ -30,9 +30,10 @@ const buildRequestPayload = (
   id: string | null,
 ): IGroupRequest | IGroupPatchRequestObject => {
   // submit Class or subclass if present, submit division or district if present, submit a name, submit projects
+  console.log(form)
```
millaeerola, line 33: This could maybe be removed

millaeerola:
When opening the advanced options, "create group" button is disabled cause "suurpiiri" is required. But if the advaced options are closed, it is possible to create a group without "suurpiiri". Should this work like this?
Screenshot 2023-12-18 at 8 42 14

src/components/Planning/GroupDialog/GroupDialog.tsx
```diff
@@ -230,7 +235,7 @@ const DialogContainer: FC<IDialogProps> = memo(
                         <TextField
                           {...formProps('name')}
                           rules={{
-                            required: t('validation.required', { value: 'Ryhman nimi' }) || '',
+                            required: t('validation.required', { value: 'Ryhmän nimi' }) || '',
```
millaeerola, line 238: Good find! 👍

src/components/Planning/GroupDialog/GroupProjectSearch.tsx
```diff
@@ -62,8 +64,8 @@ const GroupProjectSearch: FC<IProjectSearchProps> = ({
         (showAdvanceFields &&
           (!getValues('district')?.value ||
             (divisions.length > 0 && !getValues('division')?.value) ||
-            !getValues('subClass')?.value)) ||
-        (!showAdvanceFields && !getValues('subClass')?.value)
+            !getValues('subClass')?.value)) //||
+        //(!showAdvanceFields && !getValues('subClass')?.value)
```
millaeerola, line 68: Is this line of code, that is commented out, still needed?

saijaseppa: I'll delete that.

src/components/Planning/GroupDialog/GroupDialog.tsx
saijaseppa: Will delete!

millaeerola (approved):
💎

PR #128, opened 2023-12-20
fix issue with loading if a cell is activated but number not changed

markspl:
- on Programming view (Ohjelmointi), if cell is activated, but the number is not changed, the loading icon stays endlessly.


millaeerola (approved):
👍

PR #129, opened 2023-12-29
Fix/open and redirect pages correctly

markspl:
- open project page correctly when copy-pasting link, or opening the page on new tab/window
- remove extra console log
- redirect user after successful sign in

### DIFF
--- a/src/components/AuthGuard/AuthGuard.tsx
+++ b/src/components/AuthGuard/AuthGuard.tsx
@@ -41,10 +41,13 @@ const AuthGuard: FC = () => {
             const userPayload = userDispatch.payload as IUser;
 
             //if (!user || user.ad_groups.length === 0) {
-            if (userDispatch.type.includes("fulfilled") && userPayload.ad_groups.length === 0) {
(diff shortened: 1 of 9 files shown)


src/components/AuthGuard/AuthGuard.tsx
```diff
@@ -33,11 +34,17 @@ const AuthGuard: FC = () => {
...
+            const userDispatch = await dispatch(getUserThunk());
+            const userPayload = userDispatch.payload as IUser;
+
+            //if (!user || user.ad_groups.length === 0) {
+            if (userDispatch.type.includes("fulfilled") && userPayload.ad_groups.length === 0) {
+              // Do not redirect user to the path where tried to access
               localStorage.removeItem(INITIAL_PATH);
-              return;
+              getApiToken();
             }
```
markspl, line 48: In some reason, this `dispatch` does not affect to `user`, so every time when `user` is called it is `null`.

This is why redirecting will not work when returning to the app after signing in.

This change works, but it should use `useAppSelector(selectUser)` and not fetch the information using payload.

millaeerola (approved):
nice! 💎

PR #131, opened 2024-01-09
Relese 9.1.2024

millaeerola:
- Fix cell loading spinner
- Add possibility to add groups without sub class information if the hierarchy doesn't have sub classes


markspl (approved):
👍

PR #132, opened 2024-01-12
Change name of the option in project card

saijaseppa:
Option in project card was before "Puisto tai taitorakenne" and this was fixed to be only "Puisto". 
[HKISD-17](https://futurice.atlassian.net/browse/HKISD-17)


millaeerola (approved):
👍

PR #133, opened 2024-01-12
HKISD-12 / HKISD-26: Problem with the framebudgets

JenniLehtonen:
There were two similar tickets:
https://futurice.atlassian.net/browse/HKISD-12
https://futurice.atlassian.net/browse/HKISD-26

This fix solves the problem with the framebudgets now but later when we have time, we should check the possible problem with the projectGroup/projectClass thing that I commented to the code


millaeerola (approved):
Looks good 👍

PR #134, opened 2024-01-16
HKISD-26 / HKISD-12: Add check for the table level

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-26

I added this `!search.includes("subClass")` condition for the check so that all the numbers in the Eteläinen suurpiiri row (and  in other equivalent rows) are again visible on the subClass level. We will need to investigate this issue more as the framebudget is not available on the subClass level in the frontend as it's available on the higher level.
image


millaeerola (approved):
👍

PR #135, opened 2024-01-17
Fix: projects missing under subDivisions

millaeerola:
- added a check to get projects that have a subDivision to show under divisions in UI, since UI hierarchy doesn't show subDivisions


markspl (approved):
👍

PR #136, opened 2024-01-18
Fix projects didn't save under classes that don't have subClasses

millaeerola:
- Users were not able to save projects under classes that don't have subClasses


markspl (approved):
⚡

PR #137, opened 2024-01-18
Fix/modify district choice

saijaseppa:
### Description
When creating a new group, district ("suurpiiri") choice was not mandatory. But this caused problems if the subClass of the group was a district at the same time.   

This change will make district choice to be mandatory in cases where subClass of the group is a district. If subClass is something else, district choice is not mandatory. 

### Testing
Path where subClass is a district is: 803 Kadut ja liikenneväylät -> Uudisrakentaminen -> Eteläinen suurpiiri. When creating a group with this path, district choice is mandatory.


millaeerola (approved):
👍

PR #141, opened 2024-01-22
HKISD-33: fix redirecting to notes

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-33

I replaced the old confirm dialog logic (`usePromptConfirmOnNavigate`) with `ConfirmPrompt.` The project basics and notes view changing were handled by `TabList` and I replaced that with `ProjectDetailsForm`.


millaeerola:
Should this same logic be used when navigating from project notes to project basics, if there is something that is not saved in project notes?

millaeerola (approved):
Looks good! 👍

PR #142, opened 2024-01-23
fix navigating to new project card

millaeerola:
- navigating to new project card got broken here: https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/129
- it redirected the user to 404 when trying to create a new project
- this PR fixes the issue


markspl (approved):
⚡

PR #143, opened 2024-01-23
HKISD-26/fix frame budget in planning view

millaeerola:
- https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-26
- fixed frameBudget not showing correctly in planning view when opening a district so that it becomes the top level


saijaseppa (approved):
👍  good fix!

PR #144, opened 2024-01-23
don't update project form if another user edits different project

markspl:
- if user was creating a new project and another user saved updated project, a new project's form got updated project's information
- this feature is created to keep project form updated if two users are updating the same project at the same time, but in this case it was acting incorrectly
  - checks correctly if selected project is empty


millaeerola (approved):
nice! 💎

saijaseppa (approved):
This is super 💎

PR #145, opened 2024-01-23
HKISD-46/project card calendar

millaeerola:
- https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-46
- changed caledars to show dates until the end of 10th year from now instead of just exactly 10 years from now
- calendar component is from hds and they default it to 10 years from now, so it's not possible to have it be limitless. We could set the end date to very far future, but I don't know if it's a better solution than this, since it still wouldn't be limitless


saijaseppa (approved):
Calendar worked as supposed 👍

PR #146, opened 2024-01-25
HKISD-53/401/403 errors

millaeerola:
- ticket: https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-53
- viewers where getting a lot of errors and that should now be fixed
- can be tested by setting your role to just one role (not viewer, since we don't have that) from authSlice in fronend and setting the same role from baseViewSet in backend and changing Viewer role id/name from permission.py to be that role as well and the actual role to be something else (or comment that out completely)
- so login as coordinator for example and grant coordinator role viewer rights in the backend


src/i18n/fi.json
```diff
@@ -253,21 +253,26 @@
...
+      "postError": "Luonti epäonnistui",
+      "accessDenied": "Pääsy evätty"
     },
     "message": {
       "formSaveSuccess": "Lomake tallennettu onnistuneesti.",
       "noteDeleteSuccess": "Muistiinpano poistettu onnistuneesti",
       "notePatchSuccess": "Muistiinpano muokattu onnistuneesti",
       "Request failed with status code 400": "Pyyntö epäonnistui virhekoodilla 400",
+      "Request failed with status code 401": "Pyyntö epäonnistui virhekoodilla 401",
+      "Request failed with status code 403": "Pyyntö epäonnistui virhekoodilla 403",
```
JenniLehtonen, line 275: Is it okay that there are spaces between the words?

millaeerola: I agree that it's not the best way to name these, but these are the error messages that thunkAPI is creating when posting an error and for some reason they can't be modified directly. Another way to do this would be to have a converter for the messages that changes it to some better variable name that is then used for getting the translation.

JenniLehtonen: Okay I see, I think that we can keep it this way now and maybe come back to it later

JenniLehtonen (approved):

PR #147, opened 2024-01-26
HKISD-34: fix hashtag saving in new projects

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-34

When user was creating a new project, they weren't able to add hashtags for it because the hashtags were tried to be saved to a specific project right away but there were no project to which they could've been saved into.

I tried to save the hashtags to the same place where the unsaved values are stored but it was not possible (or at least not very easy) to do because of the structure of the code and I think that the hashtags were having some own state for them so they were not included to the other values of the form. I ended up creating a solution where the hashtags are temporarily stored into local storage when user is creating a project and when they want to save the project, the values are taken from the local storage. (I was also thinking if local storage should be changed to session storage or redux state, any opinions?)

I also noticed that the fix for the notes https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/141 caused a bug: if user tried to go to notes when editing a project, they got the modal that was asking if they want to continue but if the user tried to use some buttons e.g. in the side bar then the modal did not show up and they were just navigated to a new page. I added back this `usePromptConfirmOnNavigate.ts` file and deleted the new `ConfirmPrompt` files and the `getIsDirty` functions as they were not needed anymore.

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
@@ -49,8 +49,8 @@ interface IFormState {
 }
 
 const getHashtagsFromLocalStorage = () => {
-  const hashTagsInLocalStorage = localStorage.getItem('hashTagsToBeSaved');
-  return hashTagsInLocalStorage && JSON.parse(hashTagsInLocalStorage);
+  const hashtagsInLocalStorage = localStorage.getItem('hashtagsForNewProject');
+  return hashtagsInLocalStorage && JSON.parse(hashtagsInLocalStorage);
 }
 
 const ProjectHashTagsDialog: FC<IProjectHashTagsDialogProps> = forwardRef(
@@ -62,13 +62,13 @@ const ProjectHashTagsDialog: FC<IProjectHashTagsDialogProps> = forwardRef(
     const allHashTags = useAppSelector(selectHashTags);
     const { t } = useTranslation();
 
-    const initialHashTagsForNewProject = getHashtagsFromLocalStorage();
+    const initialHashtagsForNewProject = getHashtagsFromLocalStorage();
 
     /* Here are the 'initial' values to be saved to a project. When user creates a new project and inserts hashtags 
       for the first time, those will be stored here. If the user goes back to change the hashtags when they are still 
       creating the project and they decide not to change those after all (after changing the values in the modal), 
       these values will be set to the local storage and not the 'unsaved' values */
-    const [savedHashtags] = useState(initialHashTagsForNewProject);
+    const [savedHashtags] = useState(initialHashtagsForNewProject);
 
     const [formState, setFormState] = useState<IFormState>({
       hashTagsObject: {},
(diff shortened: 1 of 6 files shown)


src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
```diff
@@ -149,24 +173,65 @@ const ProjectHashTagsDialog: FC<IProjectHashTagsDialogProps> = forwardRef(
...
+    useEffect(() => {
+      if (projectMode === 'new' && hashTagsForSubmit) {
+        const hashtags = handleConcat(hashTagsForSubmit);
+        setHashTagsState((current) => ({
+          ...current,
+          projectHashTags: hashtags
+        }));
+      }
+      setHashtagClicked(false);
+    },[hashtagClicked])
```
millaeerola, line 199: Could the state be modified only when the hashTagDialog is closed by clicking "save", or what is the need for modifying it after every hashTagClick?

millaeerola (changes requested):

src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
```diff
@@ -149,24 +173,65 @@ const ProjectHashTagsDialog: FC<IProjectHashTagsDialogProps> = forwardRef(
...
+      if (hashTagsForSubmit.length) {
+        const existingHashtagIds = new Set(hashTagsForSubmit.map((hashtag) => hashtag.id));
+        // Filter out the hashtags that already exist in hashTagsForSubmit and after that combine the new values with the ones in hashTagsForSubmit
+        const newHashtags = hashtagsForSubmit.filter(hashtag => !existingHashtagIds.has(hashtag.id));
+        const combinedHashtags = [...hashTagsForSubmit, ...newHashtags];
+        // when user searches for hashtags and adds one of the search results, the hashtag will come twice --> uniqWith deletes all duplicate objects
+        return _.uniqWith(combinedHashtags, _.isEqual);
+      } 
+      return _.uniqWith(hashTagsForSubmit, _.isEqual);
+    }
```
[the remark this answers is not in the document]
JenniLehtonen: This logic was needed earlier but now it's outdated as we only use hashtagsForSubmit there, I didn't notice this earlier :smile: We'll need to use the uniqWith() to the hashtagsForSubmit still

src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
JenniLehtonen: This must've also be connected to the older logic since now it's not needed anymore

src/components/Project/ProjectBasics/ProjectForm/ProjectHashTags/ProjectHashTags.tsx
```diff
@@ -152,21 +169,37 @@ const ProjectHashTagsDialog: FC<IProjectHashTagsDialogProps> = forwardRef(
...
     );
 
     const handleClose = useCallback(() => {
+      if (projectMode === 'new') {
+        // If dialog is closed without saving set the initial values (values before making changes to the hashtags) back to the state
+       setHashTagsState((current) => ({
+          ...current,
+          projectHashTags: initialHashtagsForNewProject
+        }));
+      }
```
millaeerola, line 201: Is this now needed either, cause state is only modified if the user saves the hashtags?

millaeerola (approved):
Looks good! 💎

PR #148, opened 2024-01-26
HKISD-53: Add missing 'undefined' notification error title

markspl:
- e.status will return "undefined" if UI can't connect to API
  - show undefined as "Virhe"

Related: #146 
Ticket: https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-53


millaeerola (approved):
👍

PR #150, opened 2024-01-29
HKISD-33: fix

JenniLehtonen:
I noticed that the fix for the notes https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/141 caused a bug: if user tried to go to notes when editing a project, they got the modal that was asking if they want to continue but if the user tried to use some buttons e.g. in the side bar then the modal did not show up and they were just navigated to a new page. I added back this usePromptConfirmOnNavigate.ts file and deleted the new ConfirmPrompt files and the getIsDirty functions as they were not needed anymore


millaeerola (approved):
👍

PR #152, opened 2024-01-30
HKISD-48: Fix a calculating issue

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-48

When user was deleting a project, the numbers were not updated because they were cached in the backend (PR for that: https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/117). In addition to removing caching we have to download the data again after user deletes the project because otherwise the data won't be updated.

I was also thinking if we should do something else than delete the cache for the summed finances but then again e.g. adjusting the length of the cache does not really solve this issue either as the data should be updated when the project is deleted, any opinions?


millaeerola (approved):
👍

PR #154, opened 2024-02-02
HKISD-28: fix schedule change in project form

millaeerola:
- Fixed a bug where in moving the budgets to be within the project schedule when the project schedule changes the budget for new start year was set to NaN
- Related to this fix, theres also a fix in the backend. I refactored the backend code so that it's not limited to 10 years of finance years. PR: https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/118
- added instruction on how to add env variables to azure


JenniLehtonen (approved):
Works nicely!

PR #155, opened 2024-02-02
HKISD-44: fix project card saving timestamp

saijaseppa:
#### Situation before:
The text "saved + date + time" was updated in the lower left corner every time you opened the project card. Case was that even if the user had not tried to save anything yet, the timestamp was updated. This was in not synchronised with the backend or the database, which caused confusion in situations where data failed to update to the database: the timestamp still claimed the data was saved and updated the time. In addition, when starting creating a new project, the text "saved" was already in the bottom corner, although of course nothing had been saved at that point.
#### Problem:
So the crux of the problem was that there was no proper connection between the timestamp update and the database.
#### Solution:
Locig is now changed so that whenever a form is created or its data is updated and a save request is sent to the backend, a response is awaited from there. If the response is that the saving was successful, the timestamp is also updated. If saving to the database failed, the timestamp is not updated and the user is notified of the failed saving.

Before, the timestamp was always "Saved + date + time".
With these changes it will go as follows:
- creating a new project:
-before the first save, the timestamp would just show "Not saved."
-after saving, the timestamp would be "Last saved + date + time" (image)
-after a failed save, "Not saved" + error notification would be shown to the user
- modifying an existing project
-when you open the project card, the time stamp would be "Last saved + date + time" (picture)
-after a successful save, the timestamp would be the same, but the date+time only updated
-after a failed save, "Last saved + date + time" + error notification to the user would be displayed


src/components/Project/ProjectBasics/ProjectFormSidePanel/SaveIndicator.tsx
```diff
@@ -10,8 +10,22 @@ const SaveIndicator = () => {
...
+        text={isSaving ? 'Tallennetaan' : 'Ei tallennettu'}
+        weight={'light'}
+        textColor={'var(--color-white)'}
+        showLoading={isSaving}
+      />
+    </div>)
+  }
+  else {
+    return (
     <div className="mt-4 flex justify-start">
```
millaeerola, line 29: This if else could be refactored as something like:

`const updateText = projectMode === "new" ? 'Ei tallennettu' : t('savedTime', { time: updatedMoment })

  return (
    <div className="mt-4 flex justify-start">
      <CustomTag
        color={'var(--color-success)'}
        text={isSaving ? 'Tallennetaan' : updateText}
        weight={'light'}
        textColor={'var(--color-white)'}
        showLoading={isSaving}
      />
    </div>
  )`
to avoid repetition. Also, should the 'Ei tallennettu' text come from translations as well?

millaeerola (changes requested):

src/components/Project/ProjectBasics/ProjectFormSidePanel/SaveIndicator.tsx
saijaseppa: I made variables to translations, so now all all three comes from there: 'Ei tallennettu', 'Tallennetaan' and 'Viimeksi tallennettu'. Good proposal.

millaeerola (approved):
💎

PR #156, opened 2024-02-02
HKISD-27: fix project financial section data

xluisbras:
[HKISD-27 ](https://futurice.atlassian.net/browse/HKISD-27)

Loading the project from the link would not populate the financial section with correct project data.

The problem was related to classes, subclasses and masterClass arrays not being populated beforehand.  
Add dependency to useProjectForm hook to update formValues on classes, subClasses, masterClasses, districts, divisions, subDivisions updates.


millaeerola (approved):
Nice work! 👍

PR #158, opened 2024-02-06
HKISD-58: fix confirm modal visibility on project form

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-58

When user was on the project form and had edited some fields and then clicked some of these sidebar elements in order to move to a specific part of the form, the user was asked if the wanted to continue. I was able to get the value of the target url from args in usePromptConfirmOnNavigate that handles this confirmation modal but I'm not entirely sure if the args will always have the value in the future for example if we update the react-router package but I couldn't find the information form anywhere else either.

image

### DIFF
--- a/src/hooks/usePromptConfirmOnNavigate.ts
+++ b/src/hooks/usePromptConfirmOnNavigate.ts
@@ -5,6 +5,11 @@ import { useContext, useEffect } from 'react';
 import { UNSAFE_NavigationContext as NavigationContext } from 'react-router-dom';
 import useConfirmDialog from './useConfirmDialog';
 
+interface IArgs {
+  hash?: string;
+  pathname?: string;
+  search?: string;
+}
 const usePromptConfirmOnNavigate = ({
   title,
   description,
@@ -15,6 +20,10 @@ const usePromptConfirmOnNavigate = ({
   when: boolean;
 }) => {
   const { navigator } = useContext(NavigationContext);
+
+  // TODO: these could be gathered to some file with static data
+  const projectSideBarItems = ['#basics', '#status', '#schedule', '#financial', '#responsiblePersons', '#location', '#projectProgram'];
+ 
   const { isConfirmed } = useConfirmDialog();
 
   // Toggle a warning when trying to close the window if "when" is true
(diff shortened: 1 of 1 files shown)


src/hooks/usePromptConfirmOnNavigate.ts
```diff
@@ -15,6 +20,10 @@ const usePromptConfirmOnNavigate = ({
   when: boolean;
 }) => {
   const { navigator } = useContext(NavigationContext);
+
+  // TODO: these could be gathered to some file with static data
+  const projectSideBarItems = ['#basics', '#status', '#schedule', '#financial', '#responsiblePersons', '#location', '#projectProgram'];
+
```
millaeerola, line 26: If the idea was to catch if the user is trying to navigate out from the project form, would it be better to match the pathname with a regex patter? It could be something like
`const projectBasicsPathName = /^\/project\/.*\/(basics|new)$/`
It could also help in the future, cause if there's new sections added to the project form, this list would have to be updated, but the regex doesn't.

src/hooks/usePromptConfirmOnNavigate.ts
```diff
@@ -43,12 +52,16 @@ const usePromptConfirmOnNavigate = ({
 
     navigator.push = (...args: Parameters<typeof push>) => {
       const promptConfirmOnNavigate = async (args: Parameters<typeof push>) => {
-        // Await for the isConfirmed to either return true or false, depending on the users input
-        const confirm = await isConfirmed({ title, description });
-        if (confirm !== false) {
-
-          
+        const argsProperties = args[0] as IArgs;
+        if (argsProperties && argsProperties.hash && projectSideBarItems.includes(argsProperties.hash)) {
```
millaeerola, line 56: The regex would be used here. So it would be checked that argsProperties.pathname matches the regex which tells us that the user is not trying to navigate to a different url.

JenniLehtonen: I actually noticed that it's also possible to check if the hash exists as it is used to navigate to a specific spot on a page and if user tries to navigate to another page, the hash won't exist

JenniLehtonen: The regex solution would've been better than the original one but I guess that we can also go with checking the hash as then we wouldn't have any values that we should check or should remain the same

millaeerola: Yep, that sounds even better 👍

millaeerola (approved):
Looks good! 👍 Can be merged after solving conflicts

PR #159, opened 2024-02-07
HKISD-60: disappeared project

JenniLehtonen:
https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-60

The project disappeared because it belonged to some group and when the user changed its class in the project form, the group that the project belonged to didn't exist in the new class --> we will have to delete the group information from the project.

I also noticed another problem: if the project was under some district and the user changed the class, the district had to be removed also or the project would exist under that district in that new class that it was saved into even though that class didn't have any districts inside of it. 

When you test these changes, you might need to refresh the page after updating the project values as the  project views don't necessarily get updated.

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.tsx
@@ -224,6 +224,18 @@ const ProjectForm = () => {
             data = updateFinances(data, project);
           }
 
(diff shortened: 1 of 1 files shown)


JenniLehtonen:
We will have to delete the group data still from that specific project in the production environment to fix the existing data issue there

saijaseppa (approved):
💎

PR #161, opened 2024-02-09
HKISD-54: Fix for not to render the district twice

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-54

I was able to reproduce this issue by

1. going to a project form
2. chose a district as a subclass at Hankkeen kustannustiedot
3. then a bit lower on the form at Hankkeen sijainti, chose the same district as project's location
4. went back to the planning view using the page's own functions.

For me it didn't render the district twice if I changed the subclass and location values more than once on the project form before saving!

The extra row's type was districtPreview and I think that it should only exist on a subClass level and the solution is based on that assumption. I could find districtPreview rows e.g. here where they are on a subClass level:
image


millaeerola (approved):
Nice! 👍

PR #164, opened 2024-02-15
HKISD-22: fix budget saving and schedule editing from timeline

millaeerola:
- Ticket: [https://futurice.atlassian.net/browse/HKISD-22](https://futurice.atlassian.net/browse/HKISD-22)
## The problem
Project-update and finance-update events where setting the start year of the finance data for the project that was edited to the current year, even though not all user always have the start year of the timeline as current year. There were also some use effects in the frontend that set the startYear of projects to current year and not the actual start year the user has selected. The PlanningRow component also had it's own state containing project which sometimes didn't mach the global state and it caused problems.
## The solution
- added a converter to convert the finance data that the backend sends to start from the year that the user has selected and added the finance year to be send with the data when updating projects schedule
- removed the projects from PlanningRow's own state and started using the global state there as well
- edited the eslint rules to show warnings of missing dependencies from dependency arrays
## How to test
- move the timeline and edit project schedules and budget and check that it works correctly
- open another window and check that updates made on the other window are showing correctly in the other one (try setting timelines to start from different years)
- Check that after updating something and then moving the timeline, all projects are showing and not just the one that you modified


JenniLehtonen:
When I try to add more planning or construction years by clicking the small arrow, the view doesn't get updated but when I refresh the page then it updates
image

src/utils/common.ts
```diff
@@ -283,3 +283,83 @@ export const removeHoveredClassFromMonth = (month: string) => {
...
+        const updatedFinances = {
+          [convertedKey]: finances[key as keyof IProjectFinances]
+        };
+        convertedFinances = { ...convertedFinances, ...updatedFinances };
+      }
+    }
+  }
+  return convertedFinances;
+}
+
```
JenniLehtonen, line 326: The logic inside this function should also be commented.

src/utils/common.ts
```diff
@@ -283,3 +283,83 @@ export const removeHoveredClassFromMonth = (month: string) => {
...
+      const convertedNumber = num + yearDifference;
+      if (convertedNumber >= 0 && convertedNumber <= 10) {
+        const convertedKey = ("year" + convertedNumber) as keyof IClassFinances;
+        const updatedFinances = {
+          [convertedKey]: finances[key as keyof IClassFinances]
+        };
+        convertedFinances = { ...convertedFinances, ...updatedFinances };
+      }
+    }
+  }
```
JenniLehtonen, line 363: It would be good to write a comment here about what is happening in this loop e.g. what steps are taken inside this function. It would be easier to read it if the steps were commented.

JenniLehtonen (changes requested):
There are 60 linter errors that should also be fixed

src/utils/common.ts
millaeerola: I refactored this and the function below so that they would be easier to read and naming of things would explain the code better. I also added comments for the functions explaining why they are needed and what they do. Is there still some parts inside the functions where the logic is not clear enough?

JenniLehtonen: They look good now 👍🏻

PR #165, opened 2024-02-20
IO-334: fix construction program report

JenniLehtonen:
https://helsinkisolutionoffice.atlassian.net/browse/IO-334

I fixed the problems described in the ticket and refactored the PDF report codes so that the component would be reusable between the different reports


millaeerola (approved):
👍

PR #168, opened 2024-03-05
IO-338: fix numbers in the report

JenniLehtonen:
The numbers in the report should be frameBudgets instead of plannedBudgets


markspl (approved):
💎

PR #169, opened 2024-03-06
IO-337: Strategy report

millaeerola:
Ticket:[https://helsinkisolutionoffice.atlassian.net/browse/IO-337]( https://helsinkisolutionoffice.atlassian.net/browse/IO-337)

- Added a strategy report
- Before testing, remember to check that you have projects that are in construction or planning phase in 2023. Otherwise the report might look a bit empty :)
- Classes don't have the "Ylitysoikeus yhteensä" header, cause the only cell where that data would be displayed is the TA 2023 and customer said to leave that cell empty for now. I thought it would be a bit confusing to have the header but no data, so that's why it's not in the report for now.
- Page numbering is still missing from the footer cause I didn't find an easy way to implement that, and thought it might not be that important for now.


JenniLehtonen (approved):
Looks good!

PR #170, opened 2024-03-08
IO-338: fix data and styles

JenniLehtonen:
The customer wanted two changes to the report still:
- https://futurice.atlassian.net/browse/HKISD-81 these rows should be included there
-  they wanted some hierarchy to the report also so I added indentations and different font-weights to the levels


millaeerola (approved):
Looks good 👍

PR #173, opened 2024-03-20
fix: strategy report fixes

millaeerola:
- changed strategy report to start from current year instead of the previous year
- fixed wrong colouring of location classes
- removed groups from the report


markspl (approved):
💎

PR #174, opened 2024-03-20
IO-338: added fixes

JenniLehtonen:
I fixed the substring issue with the calculations and added a check for the names of classes. We should not show items that name does not start with some number or a letter and a space on the lower levels. This check is done for every class now but I checked that in forced to frame view data, every class should start either with numbers or a letter and a space in the higher levels so the check can be done to those levels too.


millaeerola (approved):
Looks good 👍

PR #175, opened 2024-03-20
I0-336: Operational environment analysis report

JenniLehtonen:


src/components/Report/PdfReports/OperationalEnvironmentAnalysisTableHeader.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const OperationalEnvironmentAnalysisTableHeader = () => {
+  const { t } = useTranslation();
+  const currentPlusYears = [4, 5, 6, 7, 8, 9, 10];
+  return (
+    <View style={styles.tableHeader}>
+        <Text style={styles.targetCell}>{t('target')}</Text>
+        <View style={styles.narrowerColumns}>
+            <Text>Kustannus-</Text>
+            <Text>ennuste</Text>
```
millaeerola, line 56: These could maybe be moved to the translations file.

src/utils/reportHelpers.ts
```diff
@@ -43,6 +50,83 @@ const getMonth = (dateStr: string): number => {
...
+    });
+  }
+
+  return totals;
+}
+
+const getPlannedBudgetsByCategories = (classItem: IPlanningRow, category: Categories) => {
+  const sums = sumFinances(classItem, category);
+  return sums;
+}
```
millaeerola, line 128: This function only returns the result of another function. Maybe it's not needed?

src/utils/reportHelpers.ts
```diff
@@ -249,40 +333,150 @@ const getRowType = (type: string) => {
...
+    const categoryKey = key as keyof typeof Categories;
+    categories.push({
+      children: [],
+      frameBudgets: [],
+      plannedBudgets: {},
+      plannedBudgetsForCategories: getPlannedBudgetsByCategories(project, Categories[categoryKey]),
+      id: `category-${key}-${project.id}`,
+      name: t(`option.${key}`),
+      projects: [],
+      type: "category",
```
millaeerola, line 349: The key is already a string, so maybe there's no need to get it from Categories? But this kind of relates to the previous question of if the category needs to be typeof Categories. If not, then the categories could be used from the store (listsSlice).

src/utils/staticData.ts
millaeerola, line 1: Instead of this we could use it from store?

millaeerola (changes requested):

src/utils/reportHelpers.ts
```diff
@@ -43,6 +50,83 @@ const getMonth = (dateStr: string): number => {
...
+    plannedInitial2: finances[5].plannedBudget,
+    plannedInitial3: finances[6].plannedBudget,
+    plannedInitial4: finances[7].plannedBudget,
+    plannedInitial5: finances[8].plannedBudget,
+    plannedInitial6: finances[9].plannedBudget,
+    plannedInitial7: finances[10].plannedBudget,
+  }
+};
+
+const sumFinances = (classItem: IPlanningRow, category: Categories, totalsParam?: ITotals) => {
```
millaeerola, line 84: Is there a reason why category needs to be typeof Categories and not string?

src/utils/reportHelpers.ts
JenniLehtonen: I need to use the `categoryKey` here because I need to pass the category to `getPlannedbudgetsByCategories` function and if I try to pass the category as a string I get a type error because that function expects the type to be  of Categories. I wrote that function to expect the category to be typeof Categories in order to have a better type check there

src/utils/staticData.ts
JenniLehtonen: I was thinking that do we need to store the categories in the redux store? There are some other values as well in the listSlice that I think we don't need to store there e.g. types, constructionPhaseDetails, projectQualitylevels etc.? We are now making database queries to get these values and I think that at least those properties that do not have much values could be stored in staticData as enums and we could use them from there. We could then also delete unnecessary database queries and also reduce the size of the redux store a bit. What do you think about that? I think that users in the infratool can't add more values to those properties through the UI so from that point of view we don't need to have those values in the database?

millaeerola: It's planned that the admin view would have the functionality to add and remove items from various lists in the future and that's why they are stored in the database. I'm not sure if it's planned that the admins would be able to change all lists or just some, so if there's some lists that the admins can't change, this would make sense. I think before making this change, we should discuss with the client about that feature.

millaeerola: https://helsinkisolutionoffice.atlassian.net/browse/IO-353
Screenshot 2024-03-25 at 9 31 24

I found this info related to the lists, but I'm not sure if it's outdated or still relevant.

JenniLehtonen: Okay I wasn't aware of those plans 👍🏻 This could be checked when these features are implemented as the specs might change

src/utils/reportHelpers.ts
JenniLehtonen: The Categories enum needs to be used here also because with the slice solution it would be difficult to get the budgets for the children items

millaeerola (approved):
👍

PR #179, opened 2024-03-26
IO-336: fix row hiding

JenniLehtonen:
The extra rows were not included to the if sentence so they were shown still.


millaeerola (approved):
👍

PR #180, opened 2024-03-27
HKISD-86: description length

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-86
The maximum length of this field is 1000 characters, which hasn't been communicated anywhere. This rule will add a notification text when the user exceeds the maximum length. We should check the other fields too if there are similar restrictions that could be communicated by using the already existing rules.
image


millaeerola (approved):
Nice!

PR #182, opened 2024-03-27
IO-334/fixes for construction report

millaeerola:
- Report now has all groups that have a costEstimateBudget that is grater than 1 million and have at least one project that is in planning or construction phase in the next three years
- It should also have projects that have a costEstimateBudget that is grater than 1 million and don't belong to a group
- changed masterClass names to be in uppercase if the report is constructed from coordination view data
- ticket: [https://helsinkisolutionoffice.atlassian.net/jira/software/c/projects/IO/boards/302?selectedIssue=IO-334](https://helsinkisolutionoffice.atlassian.net/jira/software/c/projects/IO/boards/302?selectedIssue=IO-334)

### DIFF
--- a/src/interfaces/reportInterfaces.ts
+++ b/src/interfaces/reportInterfaces.ts
@@ -39,6 +39,27 @@ export type getPlanningDataType = Promise<{
     groupRes: IGroup[];
     initialSelections: IPlanningRowSelections}>;
 
+export type IPlanningData = {
+  res: IProjectsResponse;
+  projects: IProject[];
+  classHierarchy: IClassHierarchy;
+  planningDistricts: {
+    districts: ILocation[];
+    year: number;
+    allLocations?: undefined;
+    divisions?: undefined;
+    subDivisions?: undefined;
+  } | {
+    districts: ILocation[];
+    allLocations: ILocation[];
+    divisions: ILocation[];
+    subDivisions: ILocation[];
+    year: number;
+  };
+  groupRes: IGroup[];
+  initialSelections: IPlanningRowSelections;
+}
+
 export const reports = [
   'operationalEnvironmentAnalysis',
   'strategy',
--- a/src/interfaces/reportInterfaces.ts
+++ b/src/interfaces/reportInterfaces.ts
@@ -19,26 +19,6 @@ export type getForcedToFrameDataType = Promise<{
     groupRes: IGroup[];
     initialSelections: IPlanningRowSelections}>;
 
-export type getPlanningDataType = Promise<{
-  res: IProjectsResponse;
-    projects: IProject[];
-    classHierarchy: IClassHierarchy;
-    planningDistricts: {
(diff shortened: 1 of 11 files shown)


src/interfaces/reportInterfaces.ts
```diff
@@ -19,6 +19,47 @@ export type getForcedToFrameDataType = Promise<{
...
+  } | {
+    districts: ILocation[];
+    allLocations: ILocation[];
+    divisions: ILocation[];
+    subDivisions: ILocation[];
+    year: number;
+  };
+  groupRes: IGroup[];
+  initialSelections: IPlanningRowSelections;
+}
```
JenniLehtonen, line 61: I would combine these two as they seem to have same properties. We could maybe leave only the IPlanningData and then use `Promise<IPlanningData>` in those places where we have a Promise with this type

src/utils/reportHelpers.ts
```diff
@@ -410,17 +458,39 @@ const getExtraRows = (project: IPlanningRow, categoriesFromSlice: IListItem[] |
...
+}
+
+const getGroupEndYear = (projects: IProject[]) => {
+  let latestConstructionEndYear: number | null = null;
+  const isLater = (p: IProject) => {return ((!latestConstructionEndYear && p.constructionEndYear) || (p.constructionEndYear && latestConstructionEndYear && p.constructionEndYear > latestConstructionEndYear))};
+  for (const p of projects) {
+    if (isLater(p)) {
+      latestConstructionEndYear = p.constructionEndYear
+    }
+  }
```
JenniLehtonen, line 479: Indentations could be fixed in isEarlier and isLater functions

src/utils/reportHelpers.ts
```diff
@@ -462,141 +532,48 @@ export const convertToReportRows = (coordinatorRows: IPlanningRow[], reportType:
...
-      case Reports.ConstructionProgram:
-        classParents = allClasses
-          .filter((ac) => classesForProjects.findIndex((cfp) => cfp.parent === ac.id) !== -1)
-          .map((c) => ({
-            ...initialValues,
+    case Reports.ConstructionProgram: {
+      const planningHierarchy = [];
+      for (const c of rows) {
+        if (c.type === 'group' && c.costEstimateBudget && parseFloat(c.costEstimateBudget.replace(/\s/g, '')) >= 1000) {
+          console.log(c.costEstimateBudget)
```
JenniLehtonen, line 539: This console.log could be removed

JenniLehtonen (changes requested):
The client asked me to add some indentations to the report row names in some other  reports, maybe we could add those here too? E.g. the rows under each "top-level" class (like 803 Kadut ja liikenneväylät in the image below) could have some space on the left side.

There were some duplicate rows in the report, I guess there shouldn't be? 
image

JenniLehtonen (approved):
Looks good!

PR #183, opened 2024-03-28
HKISD-105: disappearing form data

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-105

The logic of this fix is described in the comment in the code. I noticed that this problem occured also when user was editing a project and not just when they were creating one.


millaeerola (commented):
The issue with data disappearing from a new project card was fixed, but now if I have the same project card open in two tabs and I modify the other the other doesn't get updated. Does the same happen for you as well or is it just in my local environment? 😄

millaeerola (approved):
Looks good! 👍

PR #197, opened 2024-04-15
HKISD-120: fix search button

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-120

When user tried to use next or previous buttons on the search page, they got an error. I think that ensuring we use the same beginning in the search url, the problem could be fixed. If next / previous buttons were clicked, the search url started with **http** instead of **https**, which caused the error. The `req.fullPath` was used with the next / previous buttons and the `${REACT_APP_API_URL}/projects/search-results/?${req.params}&limit=${req.limit}&order=${req.order}` was used when user clicked the numbers.


millaeerola (approved):
Looks good! 👍

PR #200, opened 2024-05-03
HKISD-87/fix moving budgets creates zeros issue

millaeerola:
- Moving planningStartYear and constructionEndYear from the project card didn't move the budgets to the closest year within the new schedule -> should be fixed now
- Ticket (specs of this problem in the comments): [https://futurice.atlassian.net/browse/HKISD-87](https://futurice.atlassian.net/browse/HKISD-87)


markspl (approved):
💎

PR #205, opened 2024-05-13
HKISD-131: fix years and costEstimate calculation on construction report

millaeerola:
- changed years to be from 2025 instead of 2024
- fixed cost estimate value for projects on the report (was incorrectly budget before)


JenniLehtonen (approved):
Looks good!

PR #207, opened 2024-05-16
HKISD-131/construction program report changes

millaeerola:
- changed the logic for showing project on the report based on the main class they are under (8 01, 8 04 and 8 08 should have projects in addition to groups and groups are shown as classes there)
- removed empty classes (classes without projects or groups that would be shown on the report) from the report
- ticket: [https://futurice.atlassian.net/browse/HKISD-131](https://futurice.atlassian.net/browse/HKISD-131)


JenniLehtonen (approved):
Looks good!

PR #208, opened 2024-05-17
HKISD-131 change kustannusenuste to use budget instead of p…

millaeerola:
- kustannusennuste needs to use the budget (underlined in the screenshot) numbers instead of planned budgets
- 
Screenshot 2024-05-17 at 13 26 06


markspl (approved):
🌞

PR #209, opened 2024-05-20
fix: show an exact keur number as rounded number

markspl:
https://futurice.atlassian.net/browse/HKISD-133

On hierarchy view, `cost-estimate-budget` value on group row is displayed as a decimal number. This number should be a rounded number.

- Round a decimal value (thousands).


markspl:
Before continuing, this requires re-check about could the issue be fixed on `infraohjelmointi-api` only.

PR #211, opened 2024-05-22
Fix: HKISD-124 fix bug in construction end year changing

saijaseppa:
This PR fixes bug in yearCell. Year changing was not possible due to missing check for if constructionEndYear isDirty.


millaeerola (approved):
🚢

PR #215, opened 2024-05-24
HKISD-131: feat: add emty rows after the summary rows

millaeerola:
- added empty rows after the summary rows for the construction program report


markspl (approved):
💎

PR #218, opened 2024-05-28
HKISD-139: fix: add checks for project and groups budgets

millaeerola:
- added check for projects and groups that if they don't have any planned budgets on current year, year + 1 or year + 2, they will not be shown on the report
- ticket: [https://futurice.atlassian.net/browse/HKISD-139](https://futurice.atlassian.net/browse/HKISD-139)


saijaseppa (approved):
💎

PR #229, opened 2024-06-04
HKISD-138/katselijat muut user role

millaeerola:
- ticket: [https://futurice.atlassian.net/browse/HKISD-138](https://futurice.atlassian.net/browse/HKISD-138)
- added new user role based on the ticket
- the new katselijat_muut role should have the same permissions as the katselijat role
- for testing, both backend and frontend need to be in the feature branches and other roles need to be commented out

### DIFF
--- a/src/reducers/authSlice.ts
+++ b/src/reducers/authSlice.ts
@@ -1,5 +1,5 @@
 import { IError } from '@/interfaces/common';
-import { IUser } from '@/interfaces/userInterfaces';
+import { IUser, UserRole } from '@/interfaces/userInterfaces';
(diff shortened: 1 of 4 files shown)


src/reducers/authSlice.ts
```diff
@@ -1,5 +1,5 @@
 import { IError } from '@/interfaces/common';
-import { IUser } from '@/interfaces/userInterfaces';
+import { IUser, UserRole } from '@/interfaces/userInterfaces';
```
saijaseppa, line 2: This UserRole could be removed here as it's unused

saijaseppa (approved):
Great!

PR #233, opened 2024-06-06
fix: groups not showing on the report

millaeerola:
- fixed the issue of groups not showing on the construction report


markspl (approved):
💎

PR #234, opened 2024-06-06
fix: missing groups on construction report

millaeerola:
- fixes the issue of missing groups on the construction report


markspl (approved):
🔥🧯

PR #235, opened 2024-06-12
fix: HKISD-139 fix the issue with divisions not showing for groups

millaeerola:
- fixed the issue with groups not showing the division on the construction report


markspl (approved):
🌞

PR #236, opened 2024-06-13
HKISD-111: Fix operational environment analysis report

JenniLehtonen:
https://futurice.atlassian.net/browse/HKISD-111

There were some issues with the projects that are in the warranty phase. I wrote an explanation about the problem to the ticket. To solve the issue, I had to check which projects are in the warranty phase and then add calculation logic in order to get the budgets of those projects to be added to the sum levels for example in the image below (e.g. **8 01**, **8 01 03** and **8 01 03 01**).

image

**This can be tested in the following way:**
1. go to the coordination view and make sure that there is a project with some budget in the current year. Then you should take a screenshot in which you can see the numbers of different levels (e.g. **8 03**, **8 03 01** and **8 03 01 01**)
image

2. Go to change the project to the warranty phase
3. Go download the report and then compare the numbers of different levels in it to the ones in the screenshot. They should be the same.

In the coordination view the numbers on those levels are now smaller than in the report as the numbers in the report should now contain the budgets of the projects in the warranty phase.

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.test.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectForm.test.tsx
@@ -485,7 +485,7 @@ describe('projectForm', () => {
 
     mockedAxios.post.mockResolvedValueOnce(mockPostResponse);
 
-    let parentContainer = await findByTestId('project-form');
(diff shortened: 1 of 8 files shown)


millaeerola (approved):
👍

PR #239, opened 2024-06-24
HKISD-147/Add Kalasatama-Pasila with extra rows on construction report

markspl:
https://futurice.atlassian.net/browse/HKI-147

With this change, missing Kalasatama-Pasila rows "Muut alle 1 milj. euron investointihankkeet" and "Hankkeet yhteensä" rows will be added.


millaeerola (approved):
👍

PR #240, opened 2024-06-26
HKISD-147/Construction report empty summary rows

markspl:
The construction report shows empty extra summary rows, if classes does not have any printable projects.

This feature checks class' children and all projects. If the class, which will get summary rows, does not have printable projects, the feature skips and won't print summary rows.

image


millaeerola:
This issue is fixed a bit differently in this PR: [https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/246](https://github.com/City-of-Helsinki/infraohjelmointi-ui/pull/246) with other fixes to the construction report as well.

PR #242, opened 2024-07-05
HKISD-152/toimintasuunnitelma report changes

markspl:
- fix colours when end month is different than December, and the end year is different than the current year
- add colours for rows based on row types (class, subClass, project, group, ...)
- show previously hidden groups on the report
- show TA "raamiluku" values

To-do
- Hide rows that does have "raami" 0
  - Hiding classes that has TA value 0 requires more investigation and will not be done on this PR.

Testing
- Open Reports
- Download strategy Toimintasuunnitelma PDF and CSV files
  - Should include the same TA and TS values
  - TA values are only visible for higher classes
  - TS values for each classes and projects
- PDF file
  - Stylized rows

Ticket and conversation: https://futurice.atlassian.net/browse/HKISD-152

### DIFF
--- a/src/components/Report/PdfReports/StrategyTableHeader.tsx
+++ b/src/components/Report/PdfReports/StrategyTableHeader.tsx
@@ -18,6 +18,7 @@ const styles = StyleSheet.create({
     fontSize: '8px',
     fontWeight: 500,
     color: 'white',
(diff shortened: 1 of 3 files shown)


andbeda (approved):
LGTM ⛵

PR #249, opened 2024-08-09
HKISD-157/budget book summary report changes

markspl:
https://futurice.atlassian.net/browse/HKISD-157

To-Do:
- [x] Change title to _Helsingin kaupunki_
- [x] Add subtitle
- [x] Remove Swedish texts from table header row
- [x] Add bigger gap in footer
- [x] remove zeros

### DIFF
--- a/src/components/Report/PdfReports/DefaultReportFooter.tsx
+++ b/src/components/Report/PdfReports/DefaultReportFooter.tsx
@@ -1,0 +1,22 @@
+import { View, StyleSheet } from '@react-pdf/renderer';
+import { memo } from 'react';
+
+// Create styles (pdf docs can't be given rem for some reason)
+const styles = StyleSheet.create({
+  footer: {
+    maxWidth: '100%',
+    width: '100%',
+    marginBottom: '25px',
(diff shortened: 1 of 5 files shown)


src/components/Report/PdfReports/DefaultReportFooter.tsx
```diff
@@ -0,0 +1,22 @@
...
+    maxWidth: '100%',
+    width: '100%',
+    marginBottom: '25px',
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+  }
+});
+
+const StrategyTableFooter = () => {
```
millaeerola, line 16: Should this be called DefaultReportFooter instead?

millaeerola (changes requested):

src/components/Report/PdfReports/DefaultReportFooter.tsx
markspl: Ah, good catch! I'll rename that.

millaeerola (approved):
😎

PR #254, opened 2024-08-16
HKISD-151 Update formatting of sap numbers

andbeda:
Updated formatting of SAP-numbers

Sap-numbers are now displayed in kiloeuros (thousands of euros) in the overview.

Rounding: Numbers are rounded to the nearest thousand. Example: 9,501 euros → 10 000 euros.

Localization: Applied French formatting with spaces as thousand separators.


millaeerola (approved):
Looks good! 👍

PR #255, opened 2024-08-21
feat: IO-387/project card for viewers

millaeerola:
- ticket: [https://helsinkisolutionoffice.atlassian.net/browse/IO-387](https://helsinkisolutionoffice.atlassian.net/browse/IO-387)
- viewer role can now access project cards in a read only mode
- using search / seeing search results is also allowed for them

### How to test?
-UI and API need to be in the feature branches (API PR: [https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/162](https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/162)
- change your role to viewer only in both
-> for API, go to BaseViewSet.py file and remove all other role checks except for IsViewer from the list -> permission_classes should only have the IsAuthenticated and IsViewer
-> for UI, go to authSlice.ts and uncomment the viewer role/roles from ad_groups array, but leave all other roles commented out
- Open a project card and check that it's read only (no field is editable)
- Use the search and check that you can see search results and open project cards from there

### DIFF
--- a/src/components/AuthGuard/AuthGuard.tsx
+++ b/src/components/AuthGuard/AuthGuard.tsx
@@ -131,7 +131,7 @@ const AuthGuard: FC = () => {
       return navigate(-1);
     }
 
-    // Redirect to planning view if a viewer is trying to access anything but the planning view
+    // Redirect to planning view if a viewer is trying to access anything but the planning view or a project card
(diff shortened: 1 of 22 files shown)


andbeda (approved):
LGTM! ⛵

andbeda:
It would be helpful to include a section in the README that details the different user roles and how to switch between them

PR #256, opened 2024-08-21
HKISD-164/projects missing from construction report

millaeerola:
- all classes, groups and projects named "Esirankentaminen" were missing from the construction report, though only 801/Esirankentaminen class was supposed to be missing
- fixed by comparing the path instead of the name when removing the row from the report
- ticket: [https://futurice.atlassian.net/browse/HKISD-164](https://futurice.atlassian.net/browse/HKISD-164)


markspl (approved):
🌷

PR #258, opened 2024-08-22
Update/build dockerfile image

markspl:
- Switch from Docker Hub image to Red Hat Ecosystem Catalog image to fix issues with blocked image fetches from Docker Hub.
- Rewrite dockerfile
- Change boolean value to string, because docker-compose does not allow boolean value type

How-to test:
- Run `docker-compose up --build --force-recreate`
  - This force-re-builds the image

If it fails, it might happen because image caches. Cleaning docker builder caches requires deleting all images 
- `docker-compose down` - Stop and remove all docker-compose containers (doesn't delete volumes)
- `docker rmi -f infraohjelmointi_ui` - Remove `..._ui` image only
  - `docker rmi -f $(docker images -aq)` - CAUTION: Remove all docker images (if the command is needed)
- `docker builder prune` - Remove only docker builder caches that is linked to removed images

This PR should be merged into DEV that is on the same level with the main branch, in case dev requires a rollback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 # ===============================================
-FROM alpine:3.20.0
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4
 # ===============================================
-WORKDIR /app
-# Copy all files
-COPY . .
 
 ENV YARN_VERSION 1.22.19
+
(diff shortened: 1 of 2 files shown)


andbeda (approved):
👍

PR #266, opened 2024-09-20
Add max size for http header

saijaseppa:
Increase the size of http header 

Ticket: [HKI-184](https://futurice.atlassian.net/jira/servicedesk/projects/HKI/queues/custom/238/HKI-184 )

Helsinki-profiili made keyloak updates on 16.9.2024 and that increased the amount of groups in users AD_groups. That was causing login problems for many people. Users with too many ad_groups got "pääsy evätty" from infratool as the token was not valid. 

Helsinki-profiili advised to set http header size bigger, to allow more groups in user accounts. 

large_client_header_buffers 4   32k; is added as helsinki-profiili advises.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -8,19 +8,20 @@ events {
 }
 
 http {
-    include                 /etc/nginx/mime.types;
-    default_type            application/octet-stream;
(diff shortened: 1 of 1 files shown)


markspl (approved):
💎

PR #271, opened 2024-10-08
HKISD-188/fix strategy report timeline

millaeerola:
- fixed the logic to create the timeline for strategy report
- use year 2025 in filtering the project for the strategy report


markspl (approved):
💎

PR #272, opened 2024-10-09
Hkisd 188/fix strategy report timeline

millaeerola:
- fixed the inverted colours on the strategy report timeline
- fix getProjectPhase function to return the correct project phase
- remove unused getMonth function


markspl (approved):
💎

PR #273, opened 2024-10-09
HKISD-159/group project class data handling

millaeerola:
- fixed an issue where project was thrown out of group when location data was modified even when the class selection stayed the same


markspl (approved):
💎

PR #275, opened 2024-10-11
HKISD159/enable eri lokaatioita -location option

millaeerola:
- both groups and projects have "Eri kaupunginosia" location option in the division level
- If the option is selected for a group, it allows projects from all divisions to be added under the group


markspl (approved):
🚀

PR #277, opened 2024-10-17
HKISD-203 change strategy-report csv to use year 2025

millaeerola:
- changed year to be 2025 in the strategy report csv
- changed getting the frame budget to use year 2025


markspl (approved):
🐱

PR #279, opened 2024-10-18
HKSID-133 Fixed round cost estimate budget number to remove decimal

andbeda:
**Removed**: Fixed round cost estimate budget number to remove decimal. Original problem solved via backend.
**Added**: Validation for costForecast added, only whole numbers are now allowed
image


millaeerola (approved):
Looks good! 👍

PR #281, opened 2024-10-22
HKISD-159: fix groupDialog project search and showing projects in hierarchy

millaeerola:
- changed the logic of GroupDialog's project search so that if "Eri suurpiirejä" is selected as the district it's the same as if no district was selected
- same for "Eri kaupunginosia" but with division
- changed the logic of fetching projects for the planning view so that projects don't have to be directly under subClass to be fetched to enable showing projects with location information under subClasses when they are inside a group without location information that is under a subClass.


saijaseppa (approved):
💎

PR #282, opened 2024-10-29
fix: add a new and update existing hierarchy colours on strategy report

markspl:
https://futurice.atlassian.net/browse/HKISD-207

The strategy report has style issues. The hierarchy level "sublevel district" had an incorrect color, and the "collective sublevel" color was missing. This PR fixes the colors to be the same as on the hierarchy view.


millaeerola (approved):
Nice! 👍

PR #283, opened 2024-10-29
HKISD-208: feat add waiting for name in strategy report if project manager is not named

saijaseppa:
- If project has not named planning person ("vastuuhenkilö"), in strategy report (toimintasuunnitelma) there will be text "Odottaa nimeämistä" (waiting for naming) in project manager column in strategy report. Previously those were empty if responsible person was not named.


andbeda (changes requested):
In order to not have hard coded text in the helper-file. 
Something like this might be good instead:
` projectManager:
        p.personPlanning?.lastName ?? (t('report.strategy.projectManagerMissing') as string),`
 And then add the text in fi.json?

andbeda (approved):
LGTM

PR #285, opened 2024-10-31
fix: HKISD-213 add districtPreview and collectiveSubLevel to frameBud…

saijaseppa:
### About
This PR adds new types to frameBudgetHandler. 

Problem was that certain rows in report with these types had no frameBudget numbers in strategy report (toimintasuunnitelma). Adding types of those to frameBudgetHandler helps.

For the type "collectiveSubLevel" there needs to be check for path to start with 803, because outside that path, it's wanted that collectiveSubLevel types are not showing frameBudgetNumbers in strategy report. 

### Testing
1. Upload strategy report (toimintasuunnitelma). Find path 8 03 01 01 and under that "A Eteläinen suurpiiri". Previously it did not have frame budget numbers. Now there's number in column TA 2025. This was type of "previewDistrict".

2. From strategy report find path 8 04 01 02 and under that “Uudet liikuntapuistot” and “Uudet liikuntapaikat ja kentät”. Type of these is "collectiveSubLevel" but frameBudget should not be shown for these, because path does not start with 803. 

To check that this applies (copied from ticket): 
Uudisrakentamisen alta A-H (8 03 01 01), Perusparantaminen ja liikennejärjestelyt alta A-F (8 03 01 02), Muut investoinnit alta A-D (8 03 01 03) kohdista raamiluvut tulee löytyä.

### DIFF
--- a/src/utils/reportHelpers.ts
+++ b/src/utils/reportHelpers.ts
@@ -630,13 +630,15 @@ const getUnderMillionSummary = (rows: IConstructionProgramTableRow[]) => {
  * Shows current year cost estimated budget (TA, "raamiluku") on
  * Strategy report for high level classes only.
  */
-const frameBudgetHandler = (type: string, budgets: IPlanningCell[]) => {
-  if (['masterClass', 'class', 'subClass', 'subClassDistrict'].includes(type)) {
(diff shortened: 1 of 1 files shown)


millaeerola (approved):
☃️

PR #289, opened 2024-11-05
IO-403: add confirm dialog for updating all forced to frame data

millaeerola:
- added confirm dialog for updating all forced to frame data


markspl (approved):
🚀

PR #291, opened 2024-11-06
IO-401/strategy report from sovitettu budjetti data

markspl:
Create new report that uses Strategy report but uses "Sovittu budjetti" data.

To-Do:
- [x] Tests

Bugs:
- [ ] If both Strategy reports (e.g. PDF version) are downloaded without refresh, the second PDF uses same data as the PDF that was downloaded first.


markspl:
Changes looks good! 💎

millaeerola (approved):

PR #292, opened 2024-11-11
Release 11.11.2024

millaeerola:
- IO-403: forced to frame admin view
- IO-401: strategy report from forced to frame view
- HKISD-217: remove spaces from strategy report csv
- HKISD-218: fix forced to frame group sums


markspl (approved):
🚀

PR #294, opened 2024-11-13
feat: IO-410 refresh page after strategy report download

markspl:
Refresh the report page after downloading one of the Strategy files. With this, we make sure the new data for another Strategy report will be stored correctly to the report.


millaeerola (approved):
🚢

PR #309, opened 2024-12-19
HKISD-228 Fixed project header to disabled

andbeda:
Disabled project header status so that it can't be removed and not changed on top.


saijaseppa (commented):
I noticed that when changing the phase, text is changed but icon will not change in the header. Small thing, but could this be fixed?

andbeda:
> I noticed that when changing the phase, text is changed but icon will not change in the header. Small thing, but could this be fixed?

should be fixed now, thanks for noticing!

saijaseppa (approved):

PR #310, opened 2024-12-20
HKISD-254: Fix search remove selected words

markspl:
On Search view that opens from "Haku", you can remove a search word under "Mitä haet?" if there is only one on the list. The issue appears if the list has more than one search word.

The reason why this happened is the currect selector sees all search words as one text, and tries to find this from the `formValue` object. This fix selects correct element's text and uses it to remove the property from the object.

How to test before and after: Open Haku search view and add multiple search words. Try to remove one element by clicking the element.

image


saijaseppa (approved):
💪🏻

PR #311, opened 2024-12-30
Add thousand separator to sap values in project card

saijaseppa:
Adding spaces as thousand separators to sap values and budget value in project card. 

Using function formatNumberToContainSpaces from common.ts. Fixing also tests to use thousand separated values and found one missing case in mockGetResponseProvider.ts.


millaeerola (approved):
🚀

PR #319, opened 2025-02-05
HKISD-224: search groups and projects with same name

millaeerola:
- ticket: https://futurice.atlassian.net/browse/HKISD-224
- previously search only returned one project or group with same name even when there actually was more (no projects or groups with duplicate names where returned)
- you can test by creating multiple projects and groups with same name and see that search results has all of them when you search with the name
- you also need the API PR for testing: [https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/235](https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/235)


saijaseppa (approved):
🦚

PR #320, opened 2025-02-10
IO-412/Feat Ennuste-raportti

markspl:
https://helsinkisolutionoffice.atlassian.net/browse/IO-412

Adds new Ennuste report (PDF, CSV).

- Fetches both `coordinator` and `forcedToFrame` data
- Columns `Ennuste <year>` and `Poikkeama`
  - `Poikkeama` shows value that is calculated `TS <year>` - `Ennuste <year>`
  - When `Poikkeama` is bigger than €200k, style is changed to red background and white text


src/components/Report/PdfReports/TableRow.tsx
```diff
@@ -427,9 +441,24 @@ const getConstructionRowStyle = (rowType: string, depth: number) => {
...
+  if (type !== "project") return strategyReportStyles.budgetCell;
+
+  // If the value is over the threshold, the background for the project cell is red.
+  const THRESHOLD = 200;
+  const dValue = formattedNumberToNumber(deviationValueString);
+
+  if (dValue >= THRESHOLD) {
+    return forecastReportStyles.forecastDeviationCostOver;
+  }
+
```
saijaseppa, line 454: Should this be if (dValue <= -THRESHOLD)? Or then change the logic which number is decreased from what.

saijaseppa (changes requested):

saijaseppa (approved):
🪐

PR #324, opened 2025-02-18
HKISD-269: fix project card being cleared

millaeerola:
- ticket: [https://futurice.atlassian.net/browse/HKISD-269](https://futurice.atlassian.net/browse/HKISD-269)
- if a user was creating a new project or editing an existing one and someone else made a finance update, the project card was cleared
- can be tested by opening the infratool in two windows and editing or creating a new project but not saving and in the other window editing some budgets from the timeline


saijaseppa (approved):
💐

PR #325, opened 2025-02-20
HKSID-270: fix project managers permissions

saijaseppa:
### Description

Project manager is not allowed to delete projects. 
Also fixed permission to fields budgetOverRun (forbidden), projectQualityLevel (allowed), planningPhase (allowed) and constructionPhase (allowed).

Ticket: https://futurice.atlassian.net/jira/software/c/projects/HKISD/boards/62?selectedIssue=HKISD-270
Related PR in backend: https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/245

### Testing
Set the roles in backend: in file BaseViewSet.py modify the permission_classes so that it includes only the projectManager role. In frontend: in file AuthSlice.ts uncomment from the ad_groups the projectManager group. This will simulate the case when user has only projectManager role.


millaeerola (approved):
🌻

PR #326, opened 2025-02-20
HKISD-119: change location filter to use project card location lists

millaeerola:
- location filters in search used the locations from the hierarchy, which caused the lists to have multiple duplicates and missing options
- changed to use the same location options as the project card has
- API PR: https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/246


saijaseppa (approved):
🐳

PR #327, opened 2025-02-24
HKISD-119: fix dublicate classes search issue

millaeerola:
- ticket: [https://futurice.atlassian.net/browse/HKISD-119](https://futurice.atlassian.net/browse/HKISD-119)
- duplucate class and location options now have the parent class/location after it
- for example "Esirakentaminen" is now "Esirakentaminen (8 01 Kiinteä omaisuus)" and "Esirakentaminen (8 08 Projektialueiden infrarakentaminen)"
- location options has the same fix


markspl (approved):
💎

PR #340, opened 2025-03-19
test pagination for loading forced to frame projects

millaeerola:
- added pagination for when forced to frame projects are fetched for reports to see if it would make the api not crash cause of too large responses


saijaseppa (approved):
🌊

PR #341, opened 2025-03-19
fix urls for getting pages for projects to reports

millaeerola:
- fixed the urls to match the url patterns backend accepts
- not the cleanest solution, but will refactor if we notice this helps with the issue :)


markspl (approved):
🚗☁️

PR #359, opened 2025-04-22
IO-147: add validation to completed phase

millaeerola:
- added validation to when project is moved to completed phase, so that warrantyPhaseEnd can't be later than current date


markspl (approved):
💎

PR #360, opened 2025-04-22
 IO-221 change entity name validation to allow 80 chars instead of 30

millaeerola:
- ticket: [https://futurice.atlassian.net/browse/HKISD-221](https://futurice.atlassian.net/browse/HKISD-221)


markspl (approved):
🌳

PR #365, opened 2025-04-24
IO-449/forecast report minor changes

markspl:
- add kilo euros on header row and refactor texts
- show project phase on Forecast report instead of "r" and "s"
- show coordinator timeline on Forecast report instead of forcedToFrame dates

extra:
- fix: set correct width for split planning and customer cells
- refactor: change file and function names

https://helsinkisolutionoffice.atlassian.net/browse/IO-449


millaeerola (approved):
🌻

PR #367, opened 2025-04-28
fix: show default phase correctly when creating a new project

markspl:
When a new project is created, it shows string that is not translated because missing variable

image

This hides both icon and the phase string when phase is not selected.

image


millaeerola (approved):
Looks good! ⭐

PR #375, opened 2025-05-27
IO-450/report header kilo euros

markspl:
- adds missing kilo euro texts on table header rows
- remove unnecessary info text from forecast report footer
- on translation file move million euro text on shared report strings


millaeerola (approved):
✅

PR #384, opened 2025-06-18
IO-457: add SAP number to construction report and csv

millaeerola:
- added sap numbers to construction report
- add sap numbers to construction report and construction forecast csv:s


markspl (approved):
🏕️

PR #400, opened 2025-09-04
feat: IO-411 Enable programmer selection with default values

cyanidesayonara:
* Enable programmer dropdown in project form
* Add default programmer selection based on class hierarchy
* Add tests for programmer selection and defaults
* Filter out empty programmer ('Ei Valintaa') from selection dropdowns
* Implement location-based programmer selection based on suurpiiri field

https://helsinkisolutionoffice.atlassian.net/browse/IO-411

Screenshot 2025-09-04 094047


markohaarni (approved):
Works as expected and code looks good.

PR #406, opened 2025-09-29
feat: IO-745 notify user about expired session

markohaarni:
Show dialog to user if access token has expired. Clicking the confirm button in dialog refreshes the page.

https://helsinkisolutionoffice.atlassian.net/browse/IO-745


markohaarni:
> Sorry, I clicked merge by accident :) But everything looks good

Ok, no problem!

PR #413, opened 2025-10-07
fix: IO-755 Include completed projects in reports and views

cyanidesayonara:
Remove completed project filtering from ReportRow.tsx


src/components/Report/ReportRow.tsx
```diff
@@ -99,7 +119,9 @@ const ReportRow: FC<IReportRowProps> = ({ type }) => {
       forcedToFrame: false,
       year: year,
     });
-    const projects = res.results.filter((p) => p.phase.value !== 'proposal' && p.phase.value !== 'design' && p.phase.value !== 'completed');
+    const projects = res.results.filter(
```
cyanidesayonara, line 122: Actual fix is here

PR #469, opened 2026-02-18
feat: IO-389 Phase detail label and suspended project message

cyanidesayonara:
https://helsinkisolutionoffice.atlassian.net/browse/IO-389


src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
```diff
@@ -84,6 +84,19 @@ const ProjectStatusSection: FC<IProjectStatusSectionProps> = ({
     completedPhase,
   ] = useMemo(() => phases.map(({ value }) => value), [phases]);
 
+  // Phase details are per-phase from API; filter by current phase and convert to options for SelectField
+  const detailsForCurrentPhaseRaw = useMemo(
+    () =>
+      projectPhaseDetailsRaw.filter((d) => d.projectPhase?.id === currentPhase),
```
markohaarni, line 90: `d.projectPhase` seems to be string instead of object so this always returns empty array.

PR #472, opened 2026-02-24
Feature/IO-407 Add menu section to admin view

sarivah:


src/components/Admin/AdminMenus/AdminMenusTableActionButtons.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { t } = useTranslation();
+  const dispatch = useAppDispatch();
+  const isFirstRow = rowIndex === 0;
+  const isLastRow = rowIndex === rowLength - 1;
+
+  const onMoveAndSaveRow = useCallback(
+    async (rowId: string, direction: 'up' | 'down') => {
+      try {
+        dispatch(moveRow({ listType, rowId, direction }));
+        dispatch(saveTableOrderThunk({ listType, path }));
```
cyanidesayonara: Should wait for saveTableOrderThunk to return successfully here before actually showing the success message.

`await dispatch(saveTableOrderThunk({ listType, path })).unwrap();`

If .unwrap() is unsuccessful, it will go to `} catch {` and show the error message instead

src/components/Admin/AdminMenus/menuCardItems.ts
```diff
@@ -0,0 +1,56 @@
+import { ReorderableListType } from '@/interfaces/menuItemsInterfaces';
```
markohaarni, line 1: This file name would be good to start with lower case letter as it's not React component.

markohaarni:
I get error when trying to create new responsible person or programmer. Also editing them does not really work as those have no `value` field so the updated value is ignored.

src/components/Admin/AdminMenus/styles.css
```diff
@@ -0,0 +1,42 @@
+.admin-menus-view {
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+
+  .admin-menus-content {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
markohaarni: I was wondering should this center alignment be removed as it makes the layout look different than in Figma.

src/components/Admin/AdminMenus/AddOrEditMenuItemDialog.tsx
```diff
@@ -141,4 +133,246 @@ const AddOrEditMenuItemDialog: FC<AddOrEditMenuItemDialogProps> = ({
...
+  const onSetMenuItemName = useCallback(
+    (
+      value: 'firstName' | 'lastName' | 'email' | 'phone' | 'title',
+      e: ChangeEvent<HTMLInputElement>,
+    ) => setPersonTypeDialogValues({ ...personTypeDialogValues, [value]: e.target.value }),
+    [personTypeDialogValues],
+  );
+
+  const onSaveMenuItemChange = useCallback(async () => {
+    if (!Object.values(personTypeDialogValues).every((value) => !value) || !dialogState.listType)
```
markohaarni: There seems to be a mistake here as now the save logic runs only if all fields are empty, which is likely the opposite of intended behavior.

markohaarni (approved):

PR #475, opened 2026-03-11
fix: IO-821 Fix logic for determining default budget account in Talpa form

markohaarni:
Use project subClass name as default budget account if it is defined and starts with number, like "8 09 01 02 Kadut uudisrakentaminen", otherwise use class name.

https://helsinkisolutionoffice.atlassian.net/browse/IO-821


sarivah (approved):
Selkeä ja looginen korjaus pyyntöön 👍

PR #476, opened 2026-03-11
fix: IO-819 Fix updating hashtags shown in project form

markohaarni:
There was a problem with hashtags not always updating in project form when saving changes in hashtag dialog. Changed it so that hashtags are always kept up to date with form state.

https://helsinkisolutionoffice.atlassian.net/browse/IO-819


sarivah (approved):
Hyvältä näyttää!

PR #478, opened 2026-03-17
IO-823 Added stara procurement reason field to project form. 

sarivah:
- Added stara procurement method select field to project form.
- Added fetching reasons as options for the select field from backend.
- https://helsinkisolutionoffice.atlassian.net/jira/software/c/projects/IO/boards/302?selectedIssue=IO-823

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
@@ -74,7 +74,7 @@ const ProjectStatusSection: FC<IProjectStatusSectionProps> = ({
 
   useEffect(() => {
     if (!showStaraProcurementReason) {
-      setValue('staraProcurementReason', { value: '', label: '' });
+      setValue('staraProcurementReason', { value: '', label: '' }, { shouldDirty: true });
     }
   }, [showStaraProcurementReason, setValue]);
 
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
@@ -1,7 +1,7 @@
 import { FormSectionTitle, NumberField, SelectField } from '@/components/shared';
 import { FC, memo, useMemo, useState, useEffect, useCallback } from 'react';
 import { useOptions } from '@/hooks/useOptions';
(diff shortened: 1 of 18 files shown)


src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
```diff
@@ -37,23 +43,41 @@ const getPhaseIndexByPhaseId = (phaseId: string | undefined, phasesWithIndexes:
...
+    control,
+  ) as IOption | undefined;
+
+  const showStaraProcurementReason = watchedConstructionProcurementMethod?.label === 'Stara';
+
   const [phaseRequirements, setPhaseRequirements] = useState<Array<string>>([]);
 
+  useEffect(() => {
+    if (!showStaraProcurementReason) {
+      setValue('staraProcurementReason', { value: '', label: '' });
```
markohaarni: `setValue` here should have `shouldDirty` option true, otherwise the change will be discarded in onSubmit because only fields that are dirty are considered there.

src/forms/useProjectForm.ts
```diff
@@ -232,6 +233,9 @@ const useProjectForm = () => {
   // control,
   const { reset, watch, setValue, getValues, formState } = formMethods;
 
+  const useWatchField = (name: keyof IProjectForm, control: Control<IProjectForm>) =>
```
markohaarni, line 236: There is already `watch` returned from `useForm` which does the same as this. This of course works also, but I was thinking it maybe adds unnecessary code. Or is there some particular reason to use this?

sarivah: I did try to use the "watch" that was there already but it did not work as I needed: it did not react instantly to changes in the value field. Or at least I did not get it to work. I tried to watch for changes in the constructionProcurementReasonfield, but there was just no reaction when changing value in the field.

src/components/Project/ProjectBasics/ProjectForm/ProjectStatusSection.tsx
sarivah: True! Fixed it

src/forms/useProjectForm.ts
markohaarni: Ok. Then useWatch is good!

markohaarni (approved):

PR #480, opened 2026-03-20
fix: IO-825 Fix problem where data from wrong project might be shown in project form

markohaarni:
There was an issue with using selectedProject from Redux state that it might reference wrong project than what was viewed in project form. That would happen if first opening project A, then project B and then going back to project A. Data from project B would be shown as GET project would not be triggered again for project A as that data was already in cache because of RTK Query and so selectedProject in Redux would not be updated.

Fixed this by replacing using selectedProject in Redux state with useQuery from RTK Query so that correct project data would always be used. Also refactored POST, PATCH and DELETE project to use RTK Query.

https://helsinkisolutionoffice.atlassian.net/browse/IO-825

### DIFF
--- a/src/api/projectApi.ts
+++ b/src/api/projectApi.ts
@@ -1,4 +1,8 @@
-import { IProject } from '@/interfaces/projectInterfaces';
+import {
+  IProject,
+  IProjectPatchRequestObject,
+  IProjectPostRequestObject,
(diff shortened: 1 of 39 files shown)


sarivah (approved):
Hyvältä näyttää!

PR #485, opened 2026-03-26
fix: IO-836 Fix problems with readonly select field

markohaarni:
Select field threw an error when trying to access label of undefined selected value in readonly field.

Added guard for that and for resolving icon so that further crashes could be avoided.

https://helsinkisolutionoffice.atlassian.net/browse/IO-836


sarivah (approved):
Tested the problem with and without this fix and found that this works 👍 looks good!

PR #492, opened 2026-04-02
fix: IO-840 Fix problem where SAP costs were not shown for past years in planning view

markohaarni:
Made a change that summary of SAP costs in project rows is shown for past years instead of just current year.

Also made a change to redux to store and get SAP costs by year for projects and groups to support the change in displaying them.

https://helsinkisolutionoffice.atlassian.net/browse/IO-840

### DIFF
--- a/src/reducers/sapCostSlice.ts
+++ b/src/reducers/sapCostSlice.ts
@@ -2,13 +2,13 @@ import { IGroupSapCost, IProjectSapCost, ISapCost } from '@/interfaces/sapCostsI
 import { getSapCosts, getSapCurrentYear } from '@/services/sapCostsServices';
 import { RootState } from '@/store';
 import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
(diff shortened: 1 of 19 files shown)


cyanidesayonara (approved):

src/reducers/sapCostSlice.ts
```diff
@@ -82,35 +82,61 @@ export const sapCostsSlice = createSlice({
...
-              group_combined_costs: sapCost.group_combined_costs,
-            })),
-          'id',
-        ),
-      };
-    });
+    builder.addCase(
+      getSapCurrentYearThunk.fulfilled,
+      (state, action: PayloadAction<Array<ISapCost>>) => {
+        console.log('getSapCurrentYearThunk - action.payload:', action.payload);
```
cyanidesayonara: console.log is probably unnecessary here

markohaarni: True. Removed that.

PR #493, opened 2026-04-07
IO-843 Added possibility to delete items in admin menu lists from ui side

sarivah:
https://helsinkisolutionoffice.atlassian.net/jira/software/c/projects/IO/boards/302?selectedIssue=IO-843


markohaarni:
Otherwise looks good and works well, but noticed that when trying to edit project programmer, there is server error with message "'ProjectProgrammer' object has no attribute 'value'".

markohaarni (approved):

PR #494, opened 2026-04-09
IO-844 Hide postal code and city from new project view, if there are …

sarivah:
…no items found for given street address.

https://helsinkisolutionoffice.atlassian.net/jira/software/c/projects/IO/boards/302?selectedIssue=IO-844

### DIFF
--- a/src/components/Project/ProjectBasics/ProjectForm/ProjectInfoSection.tsx
+++ b/src/components/Project/ProjectBasics/ProjectForm/ProjectInfoSection.tsx
@@ -38,8 +38,13 @@ const ProjectInfoSection: FC<IProjectInfoSectionProps> = ({
   const typeQualifiers = useOptions('typeQualifiers');
   const { t } = useTranslation();
   const projectMode = useAppSelector(selectProjectMode);
-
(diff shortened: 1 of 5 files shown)


markohaarni:
When trying to save new project with empty postal code and city I get `400 Bad Request` because of null city and postalCode.

Screenshot 2026-04-13 at 12 28 08

sarivah:
@markohaarni 
> When trying to save new project with empty postal code and city I get `400 Bad Request` because of null city and postalCode.
> 
That is true but it's not related to my changes: I tested in dev to write address, remove postalCode and city as they are non-compulsory and I got the same error. The fields are not required: should sending them to api as null be allowed?

markohaarni:
> @markohaarni
> 
> > When trying to save new project with empty postal code and city I get `400 Bad Request` because of null city and postalCode.
> 
> That is true but it's not related to my changes: I tested in dev to write address, remove postalCode and city as they are non-compulsory and I got the same error. The fields are not required: should sending them to api as null be allowed?

Ah, so there was a bug there already. So yes, I guess either allow null in api or send empty string from UI.

markohaarni (approved):

PR #497, opened 2026-04-20
IO-787 Removed requirement for mode to be coordination in order to modify numbers

sarivah:
Enabled coordinators to modify numbers also in forceToFrame-view. Updated testing accordinly.

https://helsinkisolutionoffice.atlassian.net/browse/IO-787

### DIFF
--- a/src/components/Planning/PlanningTable/PlanningRow/PlanningCell.tsx
+++ b/src/components/Planning/PlanningTable/PlanningRow/PlanningCell.tsx
@@ -152,8 +152,8 @@ const PlanningCell: FC<IPlanningCellProps> = ({ type, id, cell, name }) => {
   );
 
   const isEditFrameBudgetDisabled = useMemo(
-    () => !isUserCoordinator(user) || forcedToFrame,
-    [forcedToFrame, mode, user],
+    () => !isUserCoordinator(user) || mode !== 'coordination',
+    [mode, user],
   );
 
   const totalBudgetTooltipType = type === 'group' ? 'group' : 'class';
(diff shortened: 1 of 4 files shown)


src/components/Planning/PlanningTable/PlanningRow/PlanningCell.tsx
```diff
@@ -131,14 +141,60 @@ const PlanningCell: FC<IPlanningCellProps> = ({ type, id, cell, name }) => {
       },
     };
 
+    const reduxViewType = forcedToFrame ? 'forcedToFrame' : 'coordination';
+
+    const updateCellState = (updatedData: IClass | ILocation) => {
```
markohaarni, line 146: Updated finance state should come from server sent event, but now there is something strange happening because that data is stale (for example if I update frameBudget from 200 to 300 the event returns the old "frameBudget": 200). It's also weird that in dev environment it works but locally it doesn't.

This solution works but I was wondering if the core issue with the events should be fixed as now this is kind of duplicate logic as in `useUpdateEvent.ts`.

markohaarni (approved):

src/components/Planning/PlanningTable/PlanningRow/PlanningCell.tsx
markohaarni: Problem with events will be fixed on another ticket.

sarivah: Ticket https://helsinkisolutionoffice.atlassian.net/browse/IO-889 was made to fix this issue so it won't be forgotten.

PR #499, opened 2026-04-27
IO-844 Fixed postal code search: if no street number is provided, lea…

sarivah:
…ve it out of the search as it is not required and automatically adding 1 for street number caused no results when that number was not found


markohaarni (approved):
Looks good!

PR #500, opened 2026-04-29
IO-411 Pre-fill Ohjelmoija from district when class has no default

cyanidesayonara:
* Extend useProjectProgrammer with getProgrammerForDistrict, which resolves the most-specific picked level (subDivision -> division -> district) against the backend-computed default programmer attached to each district option
* setDefaultProgrammerForClassHierarchy now falls back to the district chain when no class-side programmer is found and only fires when personProgramming is empty, so a manually picked programmer is never silently overwritten
* Carry computedDefaultProgrammer through listsSlice into a new IProjectDistrictOption


markohaarni (approved):
Looks good!

PR #501, opened 2026-04-30
feat: IO-854 Implement basis for construction handover form and first sections

markohaarni:
- Add components and client side routing for the form
- Add "Hankkeen nimi ja kuvaus", "Hankkeen aikataulu" and "Yhteyshenkilöt" sections
- Implement copy link button

https://helsinkisolutionoffice.atlassian.net/browse/IO-854

### DIFF
--- a/src/components/Project/ConstructionHandover/ConstructionHandoverContainer.tsx
+++ b/src/components/Project/ConstructionHandover/ConstructionHandoverContainer.tsx
@@ -1,0 +1,41 @@
+import { useTranslation } from 'react-i18next';
+import StartConstructionHandover from './StartConstructionHandover';
+import ConstructionHandoverForm from './ConstructionHandoverForm';
+import { ProjectFormSidePanel } from '../ProjectBasics/ProjectFormSidePanel';
+import useGetProject from '@/hooks/useGetProject';
+
+export default function ConstructionHandoverContainer() {
+  const { t } = useTranslation();
+  const { data: project } = useGetProject();
+  const isConstructionHandoverStarted = true; // NOTE: Placeholder, will be implemented later based on if construction handover data is available for the project
+
+  const navItems = [
+    { route: '#nameAndDescription', label: t('nav.nameAndDescription') },
+    { route: '#projectSchedule', label: t('nav.projectSchedule') },
+    { route: '#constructionHandoverContacts', label: t('nav.constructionHandoverContacts') },
+  ];
+
+  const handleStartHandover = () => {
+    // NOTE: Will be implemented later
+  };
+
+  return (
+    <div className="flex" data-testid="construction-handover-container">
+      <div className="flex w-[35%] flex-shrink-0 justify-center">
+        <ProjectFormSidePanel
+          navItems={navItems}
+          project={null}
+          showSaveIndicator={false}
+          showPwFolderLink={false}
+        />
+      </div>
+      <div className="project-form max-w-xl pr-4" data-testid="construction-handover-form">
+        {!isConstructionHandoverStarted ? (
+          <StartConstructionHandover onStartHandover={handleStartHandover} />
+        ) : (
+          <ConstructionHandoverForm project={project ?? null} />
+        )}
+      </div>
+    </div>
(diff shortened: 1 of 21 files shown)


src/components/Project/ConstructionHandover/ConstructionHandoverContainer.tsx
```diff
@@ -0,0 +1,41 @@
+import { useTranslation } from 'react-i18next';
+import StartConstructionHandover from './StartConstructionHandover';
+import ConstructionHandoverForm from './ConstructionHandoverForm';
+import { ProjectFormSidePanel } from '../ProjectBasics/ProjectFormSidePanel';
+import useGetProject from '@/hooks/useGetProject';
+
+export default function ConstructionHandoverContainer() {
```
sarivah, line 7: Quite a small thing, but around the project almost all of the components seems to be exported like
const Whatnot =()=>{
....
export default Whatnot
so maybe the same pattern could be used here? Also a lot of the components in general seem to be memoized:
export default memo(Whatnot);
Maybe it could be considered if there are benefits in memoizing these components as well? If no profit is found then obviously no need for that

src/components/ProjectDetailsForm/projectDetailsForm.tsx
```diff
@@ -49,6 +50,15 @@ const ProjectDetailsForm = ({ projectMode }: IProjectDetailsProps) => {
             {t('talpa')}
           </button>
         )}
+        {projectMode !== 'new' && !isOnlyViewer && (
+          <button
+            role="link"
+            className={onConstructionHandoverPage ? 'buttonHighlighted' : 'button'}
+            onClick={() => navigate('construction-handover')}
+          >
```
sarivah, line 45: The sonar error could be fixed, maybe there might be a solution in hds for this..? A button that accessibly acts as a link or something?

src/utils/common.ts
```diff
@@ -68,6 +69,11 @@ export const groupOptions = <T>(
     return groups;
   }, []);
 
+export const personToOption = (person?: IPerson): IOption => ({
+  label: person ? `${person.firstName} ${person.lastName}` : '',
+  value: person ? person.id : '',
+});
+
```
sarivah, line 76: I know this is just moved and not modified by this request, but still, would this perhaps be better something like this:

export const personToOption = (person?: IPerson): IOption | undefined => {
  return person ? {label: ${person.firstName} ${person.lastName}`, value: person.id } : undefined
}
And then filter out the undefineds..? I think we don't want options that are like {label: "", value: ""}

sarivah:
Just a couple of comments, other than that both code and the ui as a result look good! 👍

src/components/Project/ConstructionHandover/ConstructionHandoverContainer.tsx
markohaarni: Added memo() to components that take props.

I prefer defining functions including components with function keyword although I know it differs from a lot of the codebase. I just find them more readable.

src/components/ProjectDetailsForm/projectDetailsForm.tsx
markohaarni: Fixed this by using links instead of buttons. HDS Tab component would be good for this otherwise but it does not support routing.

src/utils/common.ts
markohaarni: Currently {label: "", value: ""} is considered as empty option at least in project form, so that would require a larger change. But I have also thought that it seems a bit strange that it's been implemented like that.

src/components/ProjectDetailsForm/projectDetailsForm.tsx
sarivah: Good fix!

src/utils/common.ts
sarivah: Okay, I don't think we want to make any larger changes now, so let's just go with this :)

sarivah (approved):
Looks great and good fixes too!

PR #502, opened 2026-05-04
IO-844 Refactored finding postal code to look for the first item in t…

sarivah:
…he list that definitely has the code


markohaarni (approved):
Looks good!

PR #503, opened 2026-05-04
IO-877 Made title and phone number into required fields. Fixed menu i…

sarivah:
…tem dialog validation alltogether


markohaarni (approved):
Code looks good and works as expected!

PR #504, opened 2026-05-04
IO-865 Show localized toast when PW-hanketunnus is not found in PW

cyanidesayonara:
* Add getProjectPatchErrorMessage helper that maps {"hkrId": ["PW_PROJECT_NOT_FOUND"]} to "pwProjectNotFound" toast, falls back to "formSaveError" for all other failures
* Add helper into ProjectForm and ProjectHeader (both use patchProject and both were showing the generic toast)
* Add "pwProjectNotFound" i18n key to fi.json

### DIFF
--- a/src/utils/projectErrorMessage.ts
+++ b/src/utils/projectErrorMessage.ts
@@ -1,0 +1,24 @@
+// IO-865: pick the toast for a patchProject failure. The API marks the
+// "PW-hanketunnusta ei löydy" case with a 400 body of
+// `{ hkrId: ["PW_PROJECT_NOT_FOUND"] }`; everything else falls through to
+// the generic formSaveError. RTK baseQuery wraps the axios error so the
+// response body lives at `error.data`, not `error.response.data`.
+
+export const PW_PROJECT_NOT_FOUND_CODE = 'PW_PROJECT_NOT_FOUND';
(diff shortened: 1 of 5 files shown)


src/utils/projectErrorMessage.ts
```diff
@@ -0,0 +1,24 @@
...
+export type ProjectPatchErrorMessageKey = 'pwProjectNotFound' | 'formSaveError';
+
+export const getProjectPatchErrorMessage = (
+  error: unknown,
+): ProjectPatchErrorMessageKey => {
+  if (error && typeof error === 'object') {
+    const data = (error as { data?: unknown }).data;
+    if (data && typeof data === 'object') {
+      const hkrId = (data as { hkrId?: unknown }).hkrId;
+      if (Array.isArray(hkrId) && hkrId[0] === PW_PROJECT_NOT_FOUND_CODE) {
```
sarivah: Reviewed this manually and with copilot: is it sure that if there is PW_PROJECT_NOT_FOUND_CODE present, it is the first item in array? If yes then this is okay, but if there can be multiple errors and this might not be the first one (and if it is the most essential to show to user), then it might be more beneficial to loop through the array instead of selecting the first item :)

PR #505, opened 2026-05-07
IO-878 Sync add/edit/delete of raw responsible persons/programmers to derived select lists and handle deletes in reducer. 

sarivah:
Add regression test.


markohaarni (approved):
Looks good!

PR #507, opened 2026-05-15
feat: IO-858 Implement getting, creating and updating Construction handover draft

markohaarni:
If construction handover for project does not exist, create handover button is displayed and pressing it calls POST /construction-handovers/ and form is then displayed.

Implemented save draft button which submits the form, calling PATCH /construction-handovers/{id}.

https://helsinkisolutionoffice.atlassian.net/browse/IO-858

API PR (to test this against): https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/446

### DIFF
--- a/src/components/ProjectDetailsForm/style.css
+++ b/src/components/ProjectDetailsForm/style.css
@@ -12,10 +12,10 @@
   display: inline-block;
   padding: 11px 27px;
   color: var(--color-bus) !important;
-  text-decoration: none;
+  text-decoration: none !important;
 }
 
 .buttonHighlighted {
-  border: none;
+  border: none !important;
   border-bottom: 5px solid var(--color-bus) !important;
 }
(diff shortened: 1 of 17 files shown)


src/components/ProjectDetailsForm/style.css
```diff
@@ -12,10 +12,10 @@
   display: inline-block;
   padding: 11px 27px;
   color: var(--color-bus) !important;
-  text-decoration: none;
+  text-decoration: none !important;
```
sarivah, line 15: I must ask: have we tried to use all other methods of assigning styles before using !important?

sarivah (approved):
I would check the !importants in css-files but other than that I see no problems 👍

src/components/ProjectDetailsForm/style.css
markohaarni: I know !important is not usually the best way, but in this case I thought it would be okay as a simple way, as otherwise HDS styles for link overwrite these.

sarivah: Okay, as long as it's carefully considered then it's probably fine :)

PR #511, opened 2026-05-25
feat: IO-859 Add submit to programmer and submit to construction buttons to construction handover form

markohaarni:
Also disable the form if handover status is not DRAFT.

API PR (to test against): https://github.com/City-of-Helsinki/infraohjelmointi-api/pull/450

### DIFF
--- a/src/api/constructionHandoverApi.ts
+++ b/src/api/constructionHandoverApi.ts
@@ -3,6 +3,7 @@ import {
   ConstructionHandoverStatus,
   IConstructionHandover,
   IConstructionHandoverPatchRequest,
+  IConstructionHandoverTransitionResponse,
 } from '@/interfaces/constructionHandoverInterfaces';
 import { notifySuccess } from '@/reducers/notificationSlice';
 import { t } from 'i18next';
@@ -56,7 +57,7 @@ export const constructionHandoverApi = infraohjelmointiApi.injectEndpoints({
       invalidatesTags: ['ConstructionHandovers'],
     }),
     transitionConstructionHandoverStatus: build.mutation<
-      IConstructionHandover,
+      IConstructionHandoverTransitionResponse,
       { id: string; to: ConstructionHandoverStatus }
     >({
       query: ({ id, to }) => ({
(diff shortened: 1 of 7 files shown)


sarivah:
Looks good! I'll test with the api, but before that a comment that github copilot review noticed:
Transition response type mismatch
: In constructionHandoverApi.ts:58, transitionConstructionHandoverStatus is typed to return [IConstructionHandover], but backend returns { currentStatus, possibleTransitions }. This is unsafe typing and can cause runtime bugs if mutation result is used later as a full handover object.
Fix suggestions:

- Change mutation response type to a dedicated interface, for example:
 [IConstructionHandoverTransitionResponse { currentStatus: ConstructionHandoverStatus; possibleTransitions: ConstructionHandoverStatus[] }]

Update endpoint typing in [constructionHandoverApi.ts:58]) to use that interface.

Keep [invalidatesTags: ['ConstructionHandovers']] so UI still refreshes full handover data from query after transition.

Add/adjust test to assert expected response shape for the transition mutation hook.

markohaarni:
> Looks good! I'll test with the api, but before that a comment that github copilot review noticed: Transition response type mismatch : In constructionHandoverApi.ts:58, transitionConstructionHandoverStatus is typed to return [IConstructionHandover], but backend returns { currentStatus, possibleTransitions }. This is unsafe typing and can cause runtime bugs if mutation result is used later as a full handover object. Fix suggestions:
> 
> * Change mutation response type to a dedicated interface, for example:
>   [IConstructionHandoverTransitionResponse { currentStatus: ConstructionHandoverStatus; possibleTransitions: ConstructionHandoverStatus[] }]
> 
> Update endpoint typing in [constructionHandoverApi.ts:58]) to use that interface.
> 
> Keep [invalidatesTags: ['ConstructionHandovers']] so UI still refreshes full handover data from query after transition.
> 
> Add/adjust test to assert expected response shape for the transition mutation hook.

Good point! Fixed that.

sarivah (approved):
Looks good 👍

PR #514, opened 2026-06-02
IO-896 Use full window-year budget check for forced-to-frame report

cyanidesayonara:
* The forced-to-frame report only checked the first window year (Plus0), while planning checked all three, so projects funded only in +1/+2 were dropped
* Make checkProjectHasBudgets and checkGroupHasBudgets use the same Plus0 || Plus1 || Plus2 check for ConstructionProgramForcedToFrame. Forecast stays current-year-only


markohaarni (approved):
Looks good!

PR #518, opened 2026-06-11
feat: IO-900 update info texts in AdminForcedToFrame.tsx

markohaarni:
Update info texts to reflect changed behaviour for copying planning/coordination finances to forced-to-frame view (sovitettu budjetti), in that current year forced-to-frame finances are not updated.

Almost all of the code changes in AdminForcedToFrame.tsx file are from formatting (actual update in line 143).

https://helsinkisolutionoffice.atlassian.net/browse/IO-900


sarivah:
There seems to be one sonar error of a unexpected negation condition, but that seems to be an older piece of code, so I think this is fine whether you decide to fix it or not :)

sarivah (approved):

PR #528, opened 2026-07-29
feat: remove Phase and Phase-detail editing from the admin view (IO-927)

cyanidesayonara:


sarivah:
Looks good but one addition: phases and phase details are still present in type ReorderableListType (menuItemsInterfaces.ts), which is only used for the reorderable admin menus, so I think they should also be removed from that type :)

